PR #72, opened 2022-06-13
Changed layout of profile widget and fixed text overflows

Daboo1998:


src/core/utils/url.utils.ts
```diff
@@ -3,3 +3,7 @@ export const absoluteLink = (urlString: string) => {
     ? urlString
     : `https://${urlString}`;
 };
+
+export const withoutProtocol = (urlString: string) => {
+  return urlString.replace(/^https?:\/\//, '');
```
gerbenrobijn-sigbar, line 8: This only accounts for the https protocol. What about http? Sure it should be be used, but it's still there around and about:)

gerbenrobijn-sigbar (approved):

src/core/utils/url.utils.ts
Daboo1998: The `s?` in regex means that s in https is optional. So it handles both http and https

Danchev1 (approved):

PR #76, opened 2022-06-14
Added intro video to help section

Daboo1998:


src/scenes/widgets/pages/HelpWidget/components/IntroVideo/IntroVideo.styled.ts
```diff
@@ -0,0 +1,12 @@
+import YouTube from 'react-youtube';
+import styled from 'styled-components';
+
+// @ts-ignore Youtube type isn't compatible with styled-components, but works
+export const Video = styled(YouTube)`
```
gerbenrobijn-sigbar, line 5: Would it not be better to do something like`export const YoutubeContainer = styled.div` and move the import of the react-youtube to IntroVideo.tsx?

gerbenrobijn-sigbar (approved):

PR #105, opened 2022-06-24
Fix for the unity user counter when closing through the calender page

gerbenrobijn-sigbar:


Danchev1 (approved):

src/scenes/collaboration/pages/CalendarPage/CalendarPage.tsx
```diff
@@ -38,6 +39,12 @@ const CalendarPage: FC = () => {
   // TODO: make as action in store
   const leaveCollaborationSpace = () => {
     if (collaborationState.collaborationSpace) {
+      UnityService.triggerInteractionMsg?.(
```
PAKovalev, line 42: We will a few pages linked to collaboration.
Can add that functionality only to the `Collaboration.tsx`?
I mean `return` function of `useEffect`.

PAKovalev (changes requested):

src/scenes/collaboration/pages/CalendarPage/CalendarPage.tsx
gerbenrobijn-sigbar: This cannot be changed like that, because of the fly mode. The collaboration pages are not opened, but the user is still in the collaboration

PAKovalev (approved):

PR #115, opened 2022-06-30
Feature/#1841 unity service

PAKovalev:


aminbksi (approved):

src/context/Unity/types.ts
```diff
@@ -45,9 +45,9 @@ export type StageModeAction =
   | 'joined-stage'
   | 'left-stage'
   | 'kick'
-  | 'mute';
+  | 'mute';*/
```
Danchev1, line 48: commented out? can it be removed?

Danchev1 (changes requested):

src/context/Unity/types.ts
PAKovalev: Done.

Danchev1 (approved):

PR #117, opened 2022-07-01
Feature/#1848 websocket service

PAKovalev:


src/shared/hooks/usePosBusEvent.ts
```diff
@@ -0,0 +1,23 @@
...
+  event: EventKey,
+  callback: PosBusEventType[EventKey]
+) => {
+  const callbackRef = useRef(callback);
+  useEffect(() => {
+    callbackRef.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    //@ts-ignore
```
Danchev1, line 16: @ts-ignore for what? can it be removed and the test error fixed?

Danchev1 (changes requested):

Danchev1 (changes requested):

src/shared/services/unity/UnityService.tsx
```diff
@@ -85,6 +85,25 @@ export class UnityService {
...
+
+    this.unityContext?.on('RelayMessage', (target: string, message: string) => {
+      let msgJSON;
+      try {
+        msgJSON = JSON.parse(message);
+      } catch (e) {
+        console.error(`No valid JSON in relay message "${message}"`);
+        msgJSON = '';
+      }
+      try {
```
Danchev1, line 97: could this be in one try catch?

Danchev1 (changes requested):

src/shared/services/unity/UnityService.tsx
PAKovalev: One `try` for casting, second -> for processing.
Leave this for the near future. This service will be changed also.

PR #140, opened 2022-07-20
Feature/#2008 collaboration routes

PAKovalev:


src/api/repositories/magicRepository/magicRepository.api.endpoints.ts
```diff
@@ -4,7 +4,7 @@ export const magicRepositoryEndpoints = () => {
   const BASE_URL = `${appVariables.BACKEND_ENDPOINT_URL}/magic`;
 
   return {
-    base: BASE_URL,
+    fetchLink: `${BASE_URL}/:key`,
```
Daboo1998, line 7: It think it should be id or magicId instead of key (The same is in swagger)

src/scenes/magic/pages/MagicPage/MagicPage.tsx
```diff
@@ -44,73 +28,46 @@ const MagicPage: React.FC<MagicPageProps> = (props) => {
...
+
       default:
         history.replace({pathname: ROUTES.base});
         break;
     }
-  };
+  }, [history, magicStore.magic, unityStore]);
+
+  useEffect(() => {
+    magicStore.getMagicLink(key);
```
Daboo1998, line 61: It think it should be id or magicId instead of key (The same is in swagger). We need to change key to id everywhere if we use that instead.

Daboo1998 (changes requested):

src/scenes/collaboration/Collaboration.tsx
```diff
@@ -1,60 +1,82 @@
...
-import {bytesToUuid, createRoutesByConfig} from 'core/utils';
 import StageModeModalController from 'component/molucules/StageMode/StageModeModalController';
 import NewDevicePopup from 'component/popup/new-device/NewDevicePopup';
-import {ROUTES} from 'core/constants';
 
-import {PRIVATE_ROUTES} from './CollaborationRoutes';
+import {COLLABORATION_ROUTES} from './CollaborationRoutes';
 
-interface Props {}
+const Collaboration: FC = () => {
```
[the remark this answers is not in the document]
PAKovalev: I use FC all the time.

aminbksi (approved):

Daboo1998 (approved):

PR #142, opened 2022-07-22
feature(footer): styling

PAKovalev:


src/scenes/widgets/pages/WorldStatsWidget/WorldStatsWidget.tsx
```diff
@@ -8,28 +8,28 @@ import {useStore} from 'shared/hooks';
...
   const {statsDialog, worldStats} = worldStatsStore;
   const {statistics, isLoading} = worldStats;
+  const {worldId} = worldStore;
 
   const theme = useTheme();
 
   useEffect(() => {
-    if (worldStore.worldId) {
-      worldStatsStore.init(worldStore.worldId);
+    if (worldId) {
```
Daboo1998, line 25: I think it was better as it was, as then you can just cmd click on the worldId, and it will go to definition. Your way, its more complicated to go to definition (two clicks instead of one)

Daboo1998 (changes requested):

src/scenes/widgets/pages/WorldStatsWidget/WorldStatsWidget.tsx
PAKovalev: It is just field. It is not a function.

Daboo1998: But still, what was the reasoning behind changing this this way?
You can also extract worldId in the useEffect, if the reasoning was not repeating worldStore.worldId twice

PAKovalev: I was fixing another thing in this file and found this.

Daboo1998 (approved):

PR #192, opened 2022-08-18
Feature/#1885 fly around

PAKovalev:


src/scenes/collaboration/stores/RootCollaborationStore.ts
```diff
@@ -80,19 +80,18 @@ const RootCollaborationStore = types
 
       self.isModerator = isModerator;
     }),
-    leaveMeetingSpace() {
+    async leaveMeetingSpace() {
```
aminbksi, line 83: but why not function yield in store?

Daboo1998 (approved):

aminbksi (approved):

PR #197, opened 2022-08-19
Feature/debugging 2028

Daboo1998:


PAKovalev (approved):

src/stores/MainStore/models/AgoraStore/AgoraMeetingStore/AgoraMeetingStore.ts
```diff
@@ -32,8 +33,14 @@ const AgoraMeetingStore = types
   }))
   // Listeners handlers
   .actions((self) => ({
-    handleUserPublished: flow(function* (user: IAgoraRTCRemoteUser, mediaType: 'audio' | 'video') {
+    handleUserPublished: flow(function* (
+      screenShareStore: AgoraScreenShareStoreInterface,
+      user: IAgoraRTCRemoteUser,
+      mediaType: 'audio' | 'video'
+    ) {
       if ((user?.uid as string).split('|')[0] === 'ss') {
```
dmitry-yudakov, line 41: This line is unchanged but I wonder if it should be that way - maybe instead of type assertion it's better be:
`String(user.uid).split(...`

As I see user.uid is `string | number`.

Just mentioning, it's not a blocking thing

dmitry-yudakov (approved):

PR #209, opened 2022-08-25
Ability to override locally some of app config variables

dmitry-yudakov:
Useful feature for some of local testing

### DIFF
--- a/src/stores/ConfigStore/ConfigStore.ts
+++ b/src/stores/ConfigStore/ConfigStore.ts
@@ -2,7 +2,7 @@ import {types, flow} from 'mobx-state-tree';
 
 import {RequestModel} from 'core/models';
 import {api, AppConfigResponse} from 'api';
-import {appVariables} from 'api/constants';
+import {appVariables, appVariablesOverrides} from 'api/constants';
 import {AppConfigInterface} from 'api/interfaces';
 
 const ConfigStore = types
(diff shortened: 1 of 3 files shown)


PAKovalev (approved):

Daboo1998 (approved):

src/stores/ConfigStore/ConfigStore.ts
```diff
@@ -23,6 +23,11 @@ const ConfigStore = types
           appVariables[key as keyof AppConfigInterface] = value;
         });
 
+        Object.entries(appVariablesOverrides).forEach((entry) => {
+          const [key, value] = entry;
+          appVariables[key as keyof AppConfigInterface] = value as any;
```
aminbksi, line 28: why any?

dmitry-yudakov: It's a good question. I casted as `any` because of some weird TS error that didn't make a lot of sense to me. I tried a few things, but decided not to lose more time for something so simple. In the above loop `value` as `any` as well.

```
ERROR in src/stores/ConfigStore/ConfigStore.ts:28:11
TS2322: Type 'string | number | boolean' is not assignable to type 'string'.
  Type 'number' is not assignable to type 'string'.
    26 |         Object.entries(appVariablesOverrides).forEach((entry) => {
    27 |           const [key, value] = entry;
  > 28 |           appVariables[key as keyof AppConfigInterface] = value;
       |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #233, opened 2022-09-02
feature(space): leaving

PAKovalev:
#DEV-1
The Agora lose connection when user tries to leave space.
[https://momentum.nifty.pm/l/u9CurUSFiv](https://momentum.nifty.pm/l/u9CurUSFiv)

### DIFF
--- a/src/stores/RootStore.ts
+++ b/src/stores/RootStore.ts
@@ -78,6 +78,7 @@ const RootStore = types
       } catch (ex) {
         console.error('agoraStore.leave', ex);
       }
+
       try {
         yield self.collaborationStore.leave();
       } catch (ex) {
--- a/src/stores/RootStore.ts
+++ b/src/stores/RootStore.ts
@@ -66,10 +66,19 @@ const RootStore = types
       console.log('---LEAVING---');
 
       const spaceId = self.collaborationStore.space?.id || '';
-
-      yield self.mainStore.agoraStore.leave();
-      yield self.collaborationStore.leave();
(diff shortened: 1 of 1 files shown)


dmitry-yudakov (approved):

src/stores/RootStore.ts
```diff
@@ -66,11 +66,20 @@ const RootStore = types
...
-      yield self.collaborationStore.leave();
       self.meetingStore.leave(isKicked);
 
+      /*
+         FIXME: Sometimes Agora loses connection.
+         Non-reproducible issue. Leave this for now. We will remove Agora later.
+      */
+
+      try {
+        yield self.mainStore.agoraStore.leave();
```
dmitry-yudakov, line 77: If the first one fails, the second won't be called. Maybe 

```
yield Promise.all([ self.mainStore.agoraStore.leave(), self.collaborationStore.leave() ] )
```

or separate try/catch for each of them?

PAKovalev: I don't know what king of method throw this exception.

dmitry-yudakov (approved):

src/stores/RootStore.ts
```diff
@@ -66,11 +66,24 @@ const RootStore = types
...
+      } catch (ex) {
+        console.error('agoraStore.leave', ex);
+      }
```
Daboo1998, line 80: make a space between blocks

Daboo1998 (approved):

PR #236, opened 2022-09-02
Bugfix/dev 1/unity default volume

PAKovalev:
#DEV-1
Unity volume and default volume fixing.
[https://momentum.nifty.pm/l/u9CurUSFiv](https://momentum.nifty.pm/l/u9CurUSFiv)

### DIFF
--- a/src/stores/MainStore/models/UnityStore/UnityStore.ts
+++ b/src/stores/MainStore/models/UnityStore/UnityStore.ts
@@ -13,7 +13,7 @@ const UnityStore = types
     isInitialized: false,
     isTeleportReady: false,
     muted: false,
-    volume: types.optional(types.number, 0.1),
+    volume: types.optional(types.number, 0.75),
     fetchRequest: types.optional(RequestModel, {})
   })
   .volatile<{unityContext: UnityContext | null}>(() => ({
@@ -73,7 +73,7 @@ const UnityStore = types
       UnityService.resume();
     },
     setInitialVolume() {
-      UnityService.setSoundEffectVolume('0.1');
+      UnityService.setSoundEffectVolume(self.volume.toString());
     },
     setVolume(volume: number) {
       self.volume = volume;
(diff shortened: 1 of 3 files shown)


dmitry-yudakov (approved):

src/stores/MainStore/models/UnityStore/UnityStore.ts
```diff
@@ -73,41 +74,34 @@ const UnityStore = types
...
-      if (!self.muted && newVolume === 0) {
         UnityService.toggleAllSound(self.muted);
+        UnityService.setSoundEffectVolume(self.volume.toString());
       }
-      if (self.muted && newVolume > 0) {
-        UnityService.toggleAllSound(self.muted);
+    },
+    volumeChange(newVolume: number) {
+      if (newVolume === 0) {
+        this.mute();
```
Daboo1998, line 97: I would avoid using `this` if possible and use `self` instead, so that they are separated by order of calls,

So if mute is called by volume change, then mute has to be in a separate action set, that is defined before. It would be a good structure in my opinion. Only use `this` if you need to call recursively.

src/stores/MainStore/models/UnityStore/UnityStore.ts
```diff
@@ -73,41 +74,34 @@ const UnityStore = types
...
+        return;
       }
+
       self.volume = newVolume;
-      UnityService.setSoundEffectVolume(newVolume.toString());
+      self.muted = false;
+      UnityService.toggleAllSound(self.muted);
+      UnityService.setSoundEffectVolume(self.volume.toString());
```
Daboo1998, line 104: maybe consider calling `self.unmute` before line 101 instead?

Daboo1998 (changes requested):

src/stores/MainStore/models/UnityStore/UnityStore.ts
PAKovalev: Only `this` has access to `actions` at this moment.

src/stores/MainStore/models/UnityStore/UnityStore.ts
PAKovalev: Don't agree. Unmute sets default value.

Daboo1998: That's why I said to call it before line 101

Daboo1998: Ok, let's leave it like it is, it's Friday :)

Daboo1998 (approved):

PR #245, opened 2022-09-06
close button in toast messages

aminbksi:
https://momentum.nifty.pm/l/beFSN0K8JdFw

close button for toast messages that contains button

### DIFF
--- a/src/scenes/unity/pages/UnityPage/UnityPage.tsx
+++ b/src/scenes/unity/pages/UnityPage/UnityPage.tsx
@@ -104,7 +104,7 @@ const UnityPage: FC = () => {
         sendBack={() => {
           unityStore.sendHighFiveBack(senderId);
         }}
-        isCloseButton
+        showCloseButton
       />,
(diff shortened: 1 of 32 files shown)


src/scenes/unity/pages/UnityPage/UnityPage.tsx
```diff
@@ -104,6 +104,7 @@ const UnityPage: FC = () => {
         sendBack={() => {
           unityStore.sendHighFiveBack(senderId);
         }}
+        isCloseButton
```
Daboo1998, line 107: I would rename it to `showCloseButton`, cause that's what you do here right? Or are you really saying that the HighFiveContent is a close button?

Daboo1998 (approved):

src/scenes/unity/pages/UnityPage/UnityPage.tsx
aminbksi: fair, yeah I mean show the close button. I'll change it

PR #256, opened 2022-09-07
The fix for searching for yourself

Daboo1998:
Now if you search for yourself in online list, you will see yourself.

Nifty:  https://momentum.nifty.pm/l/u9CurUSFiv

### DIFF
--- a/src/core/models/OnlineUsersList/OnlineUsersList.ts
+++ b/src/core/models/OnlineUsersList/OnlineUsersList.ts
@@ -2,6 +2,7 @@ import {cast, flow, Instance, types} from 'mobx-state-tree';
 
 import {RequestModel, UserProfileModel, UserProfileModelInterface} from 'core/models';
 import {api, OnlineUsersResponse, UserSearchResponse} from 'api';
+import {SEARCH_MINIMAL_CHARACTER_COUNT} from 'core/constants';
 
 const OnlineUsersList = types
   .model('OnlineUsersList', {
@@ -41,7 +42,8 @@ const OnlineUsersList = types
       return self.usersRequest.isLoading;
     },
     filteredPeople(excludedPeopleIds: string[]): UserProfileModelInterface[] {
-      const users = self.searchQuery ? self.searchedUsers : self.users;
+      const users =
+        self.searchQuery.length >= SEARCH_MINIMAL_CHARACTER_COUNT ? self.searchedUsers : self.users;
       return users.filter((user) => !excludedPeopleIds?.includes(user.uuid));
     }
   }));
--- a/src/ui-kit/organisms/OnlineUsersList/OnlineUsersList.tsx
+++ b/src/ui-kit/organisms/OnlineUsersList/OnlineUsersList.tsx
@@ -1,6 +1,6 @@
 import React, {useEffect} from 'react';
 import {useTranslation} from 'react-i18next';
-import {observer} from 'mobx-react-lite';
+import {observer, useObserver} from 'mobx-react-lite';
 
 import {SearchInput, useDebouncedEffect} from 'ui-kit';
 import {OnlineUsersListInterface, UserProfileModelInterface} from 'core/models';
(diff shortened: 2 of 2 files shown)


src/ui-kit/organisms/OnlineUsersList/OnlineUsersList.tsx
```diff
@@ -71,17 +71,26 @@ const OnlineUsersList: React.FC<PropsInterface> = ({
...
+
+    if (invite) {
+      sortedUsers.push(...onlineUsersList.filteredPeople([...excludedPeople, profile.uuid]));
+    } else {
+      sortedUsers.push(
+        ...onlineUsersList.filteredPeople(
+          onlineUsersList.searchQuery.length < SEARCH_MINIMAL_CHARACTER_COUNT ? [profile.uuid] : []
+        )
+      );
+    }
```
dmitry-yudakov, line 93: I think this component is too smart and should be a lot dumber. `invite` has too much logic around it,`SEARCH_MINIMAL_CHARACTER_COUNT` is checked on too many places and we pay too much attention to profile.uuid. 

I'd suggest `invite` to only mean `showInviteButton` - perhaps name it this way.
`excludedPeopleId` should mean simply filtering out these ids from result - `profile.uuid` will probably be in it and doesn't need additional logic?

Not sure about SEARCH_MINIMAL_CHARACTER_COUNT - maybe check it only in the effect responsible for applying search and clearing search results otherwise.

I realise some of these problems are from before this PR, but still it could be a good moment to address them and reduce the complexity.

PAKovalev: Basically, this component is needed to be redesigned.

src/core/models/OnlineUsersList/OnlineUsersList.ts
```diff
@@ -2,34 +2,53 @@ import {cast, flow, Instance, types} from 'mobx-state-tree';
...
       );
 
       if (response) {
-        self.users = cast(response);
+        self.users = cast([
+          ...response.filter(
+            (user) => includeCurrentUser && bytesToUuid(user.id.data) === currentUserId
+          ),
+          ...response.filter((user) => bytesToUuid(user.id.data) !== currentUserId)
+        ]);
```
dmitry-yudakov, line 30: Maybe this logic could be separated to `moveUserToTop` utility function - just this one thing. It's used in 2 places and is not at all obvious so naming it should help.

And separate `includeCurrentUser` flag doesn't seem needed - we use it for invitations to space and all space members including current user will be excluded by `excludePeopleIds` in the view.

src/core/models/OnlineUsersList/OnlineUsersList.ts
```diff
@@ -2,34 +2,53 @@ import {cast, flow, Instance, types} from 'mobx-state-tree';
...
+    ) {
       const response: UserSearchResponse = yield self.usersRequest.send(api.userRepository.search, {
         q: self.searchQuery,
         worldId,
         online
       });
 
       if (response) {
-        self.searchedUsers = cast(response.results);
+        self.users = cast([
```
dmitry-yudakov, line 46: I like the idea of having just one array of users but I don't like that it's used by 2 separate functions - `fetchUsers` and `searchUsers` as the former is called every 30 sec to refresh - it seems in sync now as the store is used by only one component but maybe in general having just 1 function calling fetch or search API depending on the internal state would be better?

src/ui-kit/organisms/OnlineUsersList/OnlineUsersList.tsx
```diff
@@ -37,23 +37,36 @@ const OnlineUsersList: React.FC<PropsInterface> = ({
...
+      return;
+    }
+
     onlineUsersList.setSearchQuery('');
-    onlineUsersList.fetchUsers(worldId);
+    onlineUsersList.fetchUsers(worldId, profile.uuid, !invite);
+
     const timeInterval = setInterval(() => {
-      onlineUsersList.fetchUsers(worldId);
+      if (onlineUsersList.searchQuery.length < SEARCH_MINIMAL_CHARACTER_COUNT) {
```
dmitry-yudakov, line 63: Would it be better if this check was a view function in the store? `isSearchQueryActive` or valid or something?

PR #258, opened 2022-09-12
not init text chat in private space

aminbksi:
https://momentum.nifty.pm/l/TkSvNl6nY2

User is connected to wrong space in chat after failing to join gathering event

### DIFF
--- a/src/scenes/collaboration/Collaboration.tsx
+++ b/src/scenes/collaboration/Collaboration.tsx
@@ -57,20 +57,27 @@ const Collaboration: FC = () => {
       await rootStore.leaveMeetingSpace();
     }
 
-    rootStore.joinMeetingSpace(spaceId, false).catch((e) => {
-      if (e instanceof PrivateSpaceError) {
-        history.push(ROUTES.base);
-        toast.error(
-          <ToastContent
-            isDanger
-            showCloseButton
-            headerIconName="alert"
-            title={t('titles.alert')}
-            text={t('collaboration.spaceIsPrivate')}
-          />
-        );
-      }
-    });
+    rootStore
+      .joinMeetingSpace(spaceId, false)
+      .then(() => {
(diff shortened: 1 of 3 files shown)


src/scenes/collaboration/Collaboration.tsx
```diff
@@ -86,10 +86,16 @@ const Collaboration: FC = () => {
   }, [textChatStore.messageSent, textChatStore.textChatDialog.isOpen, textChatStore]);
 
   useEffect(() => {
-    if (agoraStore.appId && !textChatStore.isLoggedOn) {
+    if (agoraStore.appId && !textChatStore.isLoggedOn && collaborationStore.isPrivateStore) {
```
dmitry-yudakov, line 89: Wouldn't `rootStore.joinMeetingSpace(spaceId, false).then(...)` a better place for this? Instead of effect. Effects are less predictable and more confusing because of the deps list.

src/scenes/collaboration/stores/RootCollaborationStore.ts
```diff
@@ -64,6 +65,7 @@ const RootCollaborationStore = types
         throw new PrivateSpaceError();
       }
 
+      self.isPrivateStore = true;
```
dmitry-yudakov, line 68: It's a bit confusing to me - it could be a private store that you have an access to, which is normal situation. The abnormal situation in our case is that we don't have access.

PAKovalev: Are you sure it works?

src/scenes/collaboration/Collaboration.tsx
PAKovalev: It is not clear for me.

dmitry-yudakov (approved):

PAKovalev (approved):

PR #277, opened 2022-09-20
Feature/module federation with craco

Daboo1998:
Added plugin loading (now its static miro loading, but can be changed later to loading from BE config).

[Nifty task](https://momentum.nifty.pm/l/HWpnsmEdJv7)


src/core/utils/pluginLoader.utils.tsx
```diff
@@ -0,0 +1,40 @@
+import React, {Suspense, FC} from 'react';
+
+import {PluginLoaderInterface} from 'core/interfaces';
```
PAKovalev, line 3: Move this file to shared folder.

src/core/utils/pluginLoader.utils.tsx
```diff
@@ -0,0 +1,40 @@
...
+  // @ts-ignore: Required to load list based plugins, cause of previous problems
+  const factory = await window[scope].get(module);
+  return factory();
+};
+
+export const PluginLoader: FC<PluginLoaderInterface> = ({name, url, config, module = './App'}) => {
+  const {ready, failed} = useDynamicScript(module && url);
+
+  if (!module) {
+    return <h2>Not system specified</h2>;
```
PAKovalev, line 22: i18n

src/core/utils/pluginLoader.utils.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  if (failed) {
+    return <h2>Failed to load dynamic script: {url}</h2>;
+  }
+
+  const Component = React.lazy(loadComponent(name, module));
+
+  return (
+    <Suspense fallback="Loading Plugin">
```
PAKovalev, line 36: use `ErrorBoundary`

PAKovalev, line 36: i18n

src/scenes/collaboration/Collaboration.tsx
```diff
@@ -80,6 +89,21 @@ const Collaboration: FC = () => {
       });
   }, [agoraStore, history, rootStore, sessionStore, spaceId, t, textChatStore]);
 
+  useEffect(() => {
```
PAKovalev, line 92: I would suggest to define a separate component under `collaboration/components` to have all stuff related to plugins

src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  if (!space) {
+    return null;
+  }
+
+  return (
+    <styled.Inner>
+      <SpaceTopBar
+        title={space.name ?? ''}
+        // TODO: Take that from API
+        subtitle="Miro document"
```
PAKovalev, line 27: it is prop of PluginInterface

src/stores/ConfigStore/ConfigStore.ts
```diff
@@ -28,6 +28,9 @@ const ConfigStore = types
           appVariables[key as keyof AppConfigInterface] = value as any;
         });
 
+        //@ts-ignore
+        window['env'] = appVariables;
```
PAKovalev, line 32: Make a TODO.

PAKovalev (changes requested):

src/core/constants/pluginList.constant.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import {ROUTES} from './routePaths.constants';
+
+// This list later could be passed as parameters from API
+export const PLUGIN_LIST: (config: PluginConfigInterface) => PluginInterface[] = (config) => [
+  {
+    name: 'miro_plugin',
+    iconName: 'miro',
+    // TODO: Later change to remote url
+    url: appVariables.IS_DEV_ENVIRONMENT
```
PAKovalev, line 13: make helper

PAKovalev (changes requested):

src/stores/ConfigStore/ConfigStore.ts
```diff
@@ -28,6 +28,12 @@ const ConfigStore = types
           appVariables[key as keyof AppConfigInterface] = value as any;
         });
 
+        //@ts-ignore
+        window['env'] = {
```
PAKovalev, line 32: TODO.

PAKovalev (changes requested):

src/core/utils/pluginLoader.utils.tsx
Daboo1998: which file? PluginLoaderInterface or pluginLoader.utils.tsx ?

src/core/constants/pluginList.constant.ts
```diff
@@ -0,0 +1,20 @@
...
+import {PluginConfigInterface} from 'core/interfaces';
+import {PluginInterface} from 'core/interfaces/plugin.interface';
+
+import {ROUTES} from './routePaths.constants';
+
+// This list later could be passed as parameters from API
+export const PLUGIN_LIST: (config: PluginConfigInterface) => PluginInterface[] = (config) => [
+  {
+    name: 'miro_plugin',
+    iconName: 'miro',
```
dmitry-yudakov, line 11: I'd suggest:

```
{
  id: 'miro',
  title: 'Miro Board', // for top bar and everywhere else
  iconUrl: 'https://...' // not sure here how it should be stored but it won't be in our iconset
  iconName: 'icon-name-from-iconset', // we could use iconUrl or iconName, but it will probably be url unless the iconset is very rich

File: src/core/constants/pluginList.constant.ts [hunk 10]
```diff
@@ -0,0 +1,20 @@
...
+    // TODO: Later change to remote url
+    url: appVariables.IS_DEV_ENVIRONMENT
+      ? 'http://localhost:3001/remoteEntry.js'
+      : ' https://dev.odyssey.ninja/plugins/miro/remoteEntry.js',
+    path: ROUTES.collaboration.miro,
```
dmitry-yudakov, line 16: I'm not really sure about these paths - we could use like plugin id in url if the plugin extends the space tabs panel. But IMO it's not correct to allow plugin provide this info.

src/core/utils/pluginLoader.utils.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const container = window[scope];
+  // @ts-ignore: Required to load list based plugins, cause window[scope] does not produce a type
+  await container.init(__webpack_share_scopes__.default);
+  // @ts-ignore: Required to load list based plugins, cause of previous problems
+  const factory = await window[scope].get(module);
+  return factory();
+};
+
+export const PluginLoader: FC<PluginLoaderInterface> = ({name, url, config, module = './App'}) => {
+  const {ready, failed} = useDynamicScript(module && url);
```
dmitry-yudakov, line 19: I think it should be loaded a bit differently - not as Component and hook. The module will need to provide more than one view perhaps - potentially Tab view, Widget, Space config panel, World Config panel. The configs could be declarative at the beginning but in future we'll probably allow more flexibility.

src/core/utils/pluginLoader.utils.tsx
dmitry-yudakov: The plugins will probably have to provide some translation object if they're to support several languages. Right now for simplicity I'd suggest to only leave plain english text as is.

src/scenes/collaboration/Collaboration.tsx
Daboo1998: But the problem is that both navigation tabs and pages depend on that

src/core/constants/pluginList.constant.ts
Daboo1998: I would not focus on pluginList cause it will come from backend anyways

src/core/utils/pluginLoader.utils.tsx
Daboo1998: This is not part of plugin, so @PAKovalev is right

src/core/constants/pluginList.constant.ts
Daboo1998: name: 'miro_plugin' is strictly related to MF, so it has to stay that way (I can only change fields name)

dmitry-yudakov: Yeah, it's ok to have _plugin suffix for now.

src/core/utils/pluginLoader.utils.tsx
Daboo1998: on error I think deals with that and ErrorBoundary solution is not compatible with hooks

src/core/constants/pluginList.constant.ts
Daboo1998: For the icons, we would have to refactor the IconSvg component

src/core/utils/pluginLoader.utils.tsx
Daboo1998: Hmm I don't know, as you can basically use this PluginLoader to load components from different list. Widget List for example and each Widget is a component

src/scenes/collaboration/Collaboration.routes.tsx
```diff
@@ -3,67 +3,85 @@ import {generatePath, Redirect} from 'react-router-dom';
...
+      path: ROUTES.collaboration.base,
+      exact: true,
+      main: () => <Redirect to={ROUTES.collaboration.dashboard} />
+    }
+  ];
+
+  plugins.forEach((plugin) => {
+    baseRoutes.push({
+      ...plugin,
+      main: () => (
```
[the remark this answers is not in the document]
Daboo1998: To be discussed. Might be a config coming from BE or it can be som other component exposed from the plugin itself.

src/core/utils/pluginLoader.utils.tsx
Daboo1998: A plugin can also expose more than one module, it can be ./App, or ./Widget or ./TabView etc.

src/core/utils/pluginLoader.utils.tsx
PAKovalev: pluginLoader.utils.tsx

src/scenes/collaboration/Collaboration.tsx
```diff
@@ -163,11 +185,12 @@ const Collaboration: FC = () => {
           spaceId,
           agoraStore.isStageMode,
           !!agoraScreenShareStore.videoTrack,
+          plugins,
```
PAKovalev, line 188: why?)

PAKovalev (changes requested):

PAKovalev (approved):

craco.config.js
```diff
@@ -11,6 +12,42 @@ module.exports = {
...
+            singleton: true, 
+            eager: true,
+            requiredVersion: ">=5.1.3"
+          },
+          "styled-components": { 
+            singleton: true, 
+            eager: true,
+            requiredVersion: ">=5.3.5"
+          }
+        }
```
dmitry-yudakov, line 50: Maybe we should also include axios here? It will probably be pretty common for a plugin to make custom requests to some external service.

Just a thought, not blocking right now.

dmitry-yudakov (approved):

PR #292, opened 2022-10-03
Feature/mir 22/extract core as a package in the mono repo

Daboo1998:


package.json
```diff
@@ -5,11 +5,18 @@
   "private": true,
   "scripts": {
     "postinstall": "husky install",
-    "build": "yarn --cwd=packages/ui-kit build && yarn --cwd=packages/app build",
-    "build:deps": "concurrently \"yarn --cwd=packages/ui-kit build\" \"yarn --cwd=packages/core build\"",
+    "build:deps": "concurrently -n ui-ki,core \"yarn --cwd=packages/ui-kit build\" \"yarn --cwd=packages/core build\"",
+    "build:app": "yarn --cwd=packages/app build",
+    "build:miro": "yarn --cwd=packages/plugin-miro build",
```
dmitry-yudakov, line 10: `build:plugins` maybe? You can use yarn foreach command with include option: https://yarnpkg.com/cli/workspaces/foreach#options-include%20%230

package.json
```diff
@@ -5,11 +5,18 @@
...
+    "build:depsAndApp": "yarn build:deps && yarn build:app",
+    "build:depsAndMiro": "yarn build:deps && yarn build:miro",
+    "build": "yarn build:deps && concurrently -n miro, app \"yarn build:miro\" \"yarn build:app\"",
     "start:app": "yarn --cwd=packages/app start",
     "start:miro": "yarn --cwd=packages/plugin-miro start",
-    "start": "concurrently \"yarn start:app\" \"yarn start:miro\"",
+    "start": "concurrently -n app,miro \"yarn start:app\" \"yarn start:miro\"",
+    "buildDepsAndStart:app": "yarn build:deps && yarn start:app",
```
PAKovalev, line 17: do we really need to have all these scripts? 
I feel it will be enough:
`start`, `start:miro`
`build`, `build:miro`

we won't use all of scripts you added. Isn't it?

PAKovalev (changes requested):

package.json
Daboo1998: `buildDepsAndStart` for sure I will be using, so that I don't have to worry about deps and running everything separately.
start will also be handy to start when reps are already built.


`build:depsAndMiro` will be handy for deployment, the same with `build:depsAndApp`, depending on how we deploy, or those both can be removed and `build` can be left as is, or `build` removed and those both used.

For the rest, we can allow to run auld stuff separately if someone wants, however we don't have to leave them

dmitry-yudakov: I think just `build` applying for everything would be better for using in docker - later we could add more if we need more flexibility. And `build:app`, `build:deps`, `build:plugins` are for convenience and clearer view.

dmitry-yudakov (approved):

PAKovalev (approved):

PR #309, opened 2022-10-07
feature(sdk): added custom script for running/building and hidden cra…

Daboo1998:
…co config to sdk


packages/plugin-miro/package.json
```diff
@@ -58,10 +57,10 @@
...
   },
   "scripts": {
-    "start": "craco start",
-    "build": "craco build",
-    "test": "craco test",
-    "eject": "craco eject" 
+    "start": "momentum-sdk start",
+    "build": "momentum-sdk build",
+    "test": "momentum-sdk test",
+    "eject": "momentum-sdk eject"
```
dmitry-yudakov, line 63: You don't really have `eject` yet :)

packages/sdk/package.json
```diff
@@ -6,16 +6,24 @@
...
   },
+  "devDependencies": {
+    "copyfiles": "^2.4.1"
+  },
   "peerDependencies": {
     "react": "^16.14.0",
     "react-dom": "^16.14.0",
-    "axios": "^0.21.4"
+    "axios": "^0.21.4", 
+    "@craco/craco": "^6.1.1"
```
dmitry-yudakov, line 24: I don't think peer deps is the right place for that - it means the user would have to install it. 
I'd say put it to `dependencies`

dmitry-yudakov (approved):
Nice work! Approved with just couple of improvement suggestions

PAKovalev (approved):

PR #314, opened 2022-10-10
feature(sdk): Fixed loading multiple plugins to have different scopes

Daboo1998:


PAKovalev (approved):

packages/plugin_miro/public/manifest.json
```diff
@@ -1,5 +1,6 @@
 {
   "short_name": "React App",
+  "plugin_id": "plugin_miro",
```
dmitry-yudakov, line 3: I actually though of having another `manifest.json` in the root of the plugin repo with plugin related stuff - id, description, icon, etc.

This one is for CRA which is for dev only. We could even remove it I guess - I don't think it's used for web normally.

dmitry-yudakov (approved):
I generally approve except for the manifest suggestion.

packages/plugin_miro/public/manifest.json
Daboo1998: So you suggest to move it to root and remove other fields than plugin_id?

dmitry-yudakov: Yep, something like that - with fields `id, name, icon, description`.

Daboo1998:
yeah lets have it consistent, at least for now

PR #315, opened 2022-10-10
Feature/mir 40/try loading components from store instead of plugin loader

Daboo1998:


packages/app/src/scenes/collaboration/stores/CollaborationPluginsStore/models/CollaborationPlugin/CollaborationPlugin.ts
```diff
@@ -1,15 +1,44 @@
...
+          // @ts-ignore: Required to load list based plugins, no ts declaration
+          await __webpack_init_sharing__('default');
+          // @ts-ignore: Required to load list based plugins, window has no dict based declaration
+          const container = window[self.name];
+          // @ts-ignore: Required to load list based plugins, cause window[scope] does not produce a type
+          await container.init(__webpack_share_scopes__.default);
+          // @ts-ignore: Required to load list based plugins, cause of previous problems
+          const plugin = (await window[self.name].get('./Plugin'))().default;
+          // @ts-ignore: Required to load list based plugins, cause of previous problems
+          return plugin[`${PluginTypeEnum.SPACE}Extension`];
```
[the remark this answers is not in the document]
dmitry-yudakov: I'd say just return `plugin` here as Plugin Interface and indeed make it universal plugin loader.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -56,18 +66,21 @@ const CollaborationPluginPage: FC<PropsInterface> = ({plugin}) => {
...
-          props={{
-            theme,
-            isSpaceAdmin: space.isAdmin,
-            spaceId: space.id,
-            request: request,
-            renderTopBarActions
-          }}
-        />
+        {!failed && !plugin.isErrorWhileLoadingComponent ? (
+          ready && plugin.Component ? (
```
dmitry-yudakov, line 70: I think all these could be encapsulated into one store - the loading state, loaded plugin with all its parts. In Collaboration we would check for spaceExtension property of the loaded plugin.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -21,16 +23,24 @@ const CollaborationPluginPage: FC<PropsInterface> = ({plugin}) => {
   const {collaborationStore, mainStore, sessionStore, leaveMeetingSpace} = useStore();
   const {space, textChatStore} = collaborationStore;
   const {favoriteStore} = mainStore;
+  const {ready, failed} = useDynamicScript(module && plugin.url);
```
dmitry-yudakov, line 26: I think it shouldn't be a hook and its main functionality could go to the common plugin loader function/store.

dmitry-yudakov (approved):
It's a good step forward, but I'd say we should go further and move the whole loading logic into one store providing Plugin Interface object.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
Daboo1998: The only thing is that we move too much DOM/ui logic into store, and stores should be more for business logic.

dmitry-yudakov: Well, using DOM/script-tag here is just (IMO slightly ugly) implementation detail - it could stay as separate function encapsulating the logic that is called in the store. It's not that different from API call. I believe in future using some form of dynamic import should be possible and would be more appropriate.

Daboo1998: I don't think it is possible otherwise. However I don't really get what you are recommending here? Not to add script tag to head and importing the script from API call? I don't think that's possible, because of how Module federation/ Web pack works internally, it has to loaded to the window etc.

dmitry-yudakov: Yeah, I thought it'd be nicer with dynamic import but if it can be script-tag only it's fine.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -56,18 +82,21 @@ const CollaborationPluginPage: FC<PropsInterface> = ({plugin}) => {
...
-          pluginType={PluginTypeEnum.SPACE}
-          props={{
-            theme,
-            isSpaceAdmin: space.isAdmin,
-            spaceId: space.id,
-            request: request,
-            renderTopBarActions
-          }}
-        />
+        {!pluginLoader.isErrorWhileLoadingComponent ? (
```
dmitry-yudakov, line 85: I think it's false at the beginning, maybe just check `!!plugin`?

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -1,36 +1,62 @@
...
+  useEffect(() => {
+    pluginLoader.init();
+
+    return () => {
+      pluginLoader.deinit();
+    };
+  }, [pluginLoader]);
+
+  useEffect(() => {
+    if (pluginLoader.isErrorWhileLoadingDynamicScript) {
```
dmitry-yudakov, line 47: just `isError`?

dmitry-yudakov (approved):
it's good, 2 small remarks

PAKovalev (approved):

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
Daboo1998: No, cause if it is false then it shows that there is an error. If I just check `!!plugin` then the message won't match the condition

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
Daboo1998: will add a view which will check both errors. We need to differentiate between the errors

PR #323, opened 2022-10-11
screen share start redirect

aminbksi:
https://momentum.nifty.pm/l/gYH2thldes

Automatically redirecting users in space when screen sharing starts

### DIFF
--- a/src/scenes/meeting/pages/PosBusEventsPage/PosBusEventsPage.tsx
+++ b/src/scenes/meeting/pages/PosBusEventsPage/PosBusEventsPage.tsx
@@ -9,6 +9,7 @@ import {ROUTES} from 'core/constants';
 import {LiveStreamInterface} from 'api';
 import {usePosBusEvent, useStore} from 'shared/hooks';
 import {ToastContent, TOAST_COMMON_OPTIONS, TOAST_GROUND_OPTIONS} from 'ui-kit';
+import {PosBusScreenShareMessageType} from 'core/types';
 
 const PosBusEventsPage: FC = () => {
   const rootStore = useStore();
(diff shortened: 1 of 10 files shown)


src/stores/MainStore/models/AgoraStore/AgoraScreenShareStore/AgoraScreenShareStore.ts
```diff
@@ -40,6 +41,12 @@ const AgoraScreenShareStore = types
     },
     set videoTrack(track: IRemoteVideoTrack | undefined) {
       self._videoTrack = track;
+
+      if (self.isScreenSharing) {
+        self.request.send(api.agoraRepository.relayScreenShare, {
+          spaceId: self.spaceId
+        });
+      }
```
dmitry-yudakov, line 49: The setter somehow doesn't feel very appropriate place for sending request. Maybe somewhere in `startScreenSharing`?

src/scenes/meeting/pages/PosBusEventsPage/PosBusEventsPage.tsx
```diff
@@ -231,6 +232,13 @@ const PosBusEventsPage: FC = () => {
     }
   });
 
+  usePosBusEvent('screen-share', (message: PosBusScreenShareMessageType) => {
+    console.info('[POSBUS EVENT] screen-share', message);
+    const {spaceId} = message;
+
+    history.push(generatePath(ROUTES.collaboration.screenShare, {spaceId}));
```
dmitry-yudakov, line 239: It would be great if we could set globally some `NO_REDIRECT` state in cases like dirty form when creating/editing event for example. @PAKovalev @Daboo1998 

I think we agreed to skip it here to move it quicker to production but something to think about in future.

dmitry-yudakov (approved):
approved with small remarks

src/stores/MainStore/models/AgoraStore/AgoraScreenShareStore/AgoraScreenShareStore.ts
aminbksi: Well, the thing is in start screen sharing we already call an EP, but what we want is when the video get streamed.

PR #332, opened 2022-10-13
Feature/dev 99/resume music when leave space

aminbksi:
https://momentum.nifty.pm/l/ZqFOnLaObE
Music should resume automatically after leaving a space.

https://momentum.nifty.pm/l/87bck2vXH5
Add scrollable left column on Admin panel

### DIFF
--- a/packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
+++ b/packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
@@ -30,9 +30,7 @@ const MeetingRoomPage: FC<PropsInterface> = ({onLeave}) => {
   const location = useLocation();
 
   useEffect(() => {
-    if (musicPlayerStore.musicPlayer.isPlaying) {
-      musicPlayerStore.togglePlayback();
-    }
+    musicPlayerStore.togglePlaybackInSpace(agoraStore.hasJoined);
   }, [agoraStore.hasJoined, musicPlayerStore]);
 
   const isButtonsAvailable = useMemo(() => {
--- a/packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
+++ b/packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
@@ -9,7 +9,9 @@ const MusicPlayerStore = types.compose(
     .model('MusicPlayerStore', {
       musicPlayerWidget: types.optional(Dialog, {}),
       playlist: types.optional(Playlist, {}),
-      musicPlayer: types.optional(MusicPlayer, {})
+      musicPlayer: types.optional(MusicPlayer, {}),
+      isPausedInSpace: types.optional(types.boolean, false),
+      isInSpace: false
     })
     .actions((self) => ({
       init(worldId: string): void {
@@ -18,6 +20,9 @@ const MusicPlayerStore = types.compose(
       togglePlayback(): void {
         if (self.playlist.tracks.length < 1) {
           return;
+        }
+        if (self.isInSpace && self.musicPlayer.isPlaying && !self.isPausedInSpace) {
+          self.isPausedInSpace = true;
         }
         self.musicPlayer.isPlaying = !self.musicPlayer.isPlaying;
         if (!self.musicPlayer.isPlaying) {
@@ -76,6 +81,24 @@ const MusicPlayerStore = types.compose(
           self.playlist.first();
           self.musicPlayer.stoppedPlaying();
         }
+      },
+      togglePlaybackInSpace(joined: boolean) {
+        if (self.musicPlayer.isPlaying && joined) {
+          this.togglePlayback();
+          this.enteredSpace();
+        } else if (!self.musicPlayer.isPlaying && self.isInSpace && !self.isPausedInSpace) {
+          this.togglePlayback();
+          this.leftSpace();
+        } else if (!joined) {
+          this.leftSpace();
+        }
+      },
+      enteredSpace() {
+        self.isInSpace = true;
+      },
+      leftSpace() {
+        self.isInSpace = false;
+        self.isPausedInSpace = false;
       }
     }))
 );
--- a/packages/app/src/scenes/spaceAdmin/pages/SpaceAdminPage/SpaceAdminPage.styled.ts
+++ b/packages/app/src/scenes/spaceAdmin/pages/SpaceAdminPage/SpaceAdminPage.styled.ts
@@ -23,6 +23,7 @@ export const Body = styled.div`
 export const VerticalSplit = styled.div`
   display: flex;
   flex-direction: column;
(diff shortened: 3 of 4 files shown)


packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
```diff
@@ -30,9 +30,7 @@ const MeetingRoomPage: FC<PropsInterface> = ({onLeave}) => {
   const location = useLocation();
 
   useEffect(() => {
-    if (musicPlayerStore.musicPlayer.isPlaying) {
-      musicPlayerStore.togglePlayback();
-    }
+    musicPlayerStore.togglePlaybackInSpace(agoraStore.hasJoined);
```
dmitry-yudakov, line 33: I think it'd be a lot easier just to take the current state of player in the beginning of this effect and stop/restart playing only if it was ON. The deps list would need to be just music player and agora stores, without mentioning their properties.

packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
```diff
@@ -76,6 +81,24 @@ const MusicPlayerStore = types.compose(
...
+        } else if (!joined) {
+          this.leftSpace();
+        }
+      },
+      enteredSpace() {
+        self.isInSpace = true;
+      },
+      leftSpace() {
+        self.isInSpace = false;
+        self.isPausedInSpace = false;
```
dmitry-yudakov, line 101: I think all this is not needed. And if it is needed, then I'd suggest to use more generic terms - not "joining space".

dmitry-yudakov (changes requested):
I suggest to simplify it by putting a bit of logic in the component and not changing the store at all.

packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
aminbksi: this logic is very specific to the space

packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
dmitry-yudakov: Yeah, but it changes. We need to take a "snapshot" of it when the component is mounted and take action on unmount based on this "snapshot".

aminbksi: yes but if after entering the space you start the music, then leaving the space would cause to stop it

packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
```diff
@@ -9,7 +9,9 @@ const MusicPlayerStore = types.compose(
     .model('MusicPlayerStore', {
       musicPlayerWidget: types.optional(Dialog, {}),
       playlist: types.optional(Playlist, {}),
-      musicPlayer: types.optional(MusicPlayer, {})
+      musicPlayer: types.optional(MusicPlayer, {}),
+      isPausedInSpace: types.optional(types.boolean, false),
```
Daboo1998, line 13: Why overcomplicate yourself?
```suggestion
      isPausedInSpace: false,
```

packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
```diff
@@ -76,6 +81,24 @@ const MusicPlayerStore = types.compose(
           self.playlist.first();
           self.musicPlayer.stoppedPlaying();
         }
+      },
+      togglePlaybackInSpace(joined: boolean) {
+        if (self.musicPlayer.isPlaying && joined) {
+          this.togglePlayback();
+          this.enteredSpace();
+        } else if (!self.musicPlayer.isPlaying && self.isInSpace && !self.isPausedInSpace) {
+          this.togglePlayback();
```
Daboo1998, line 90: `togglePlayback` is misleading here. Toggling means on -> off && off -> on. While here you have only one way, so then `togglePlaybackInSpace` is also misleading

packages/app/src/scenes/widgets/stores/MusicPlayerStore/MusicPlayerStore.ts
```diff
@@ -76,6 +81,24 @@ const MusicPlayerStore = types.compose(
...
+          this.leftSpace();
+        } else if (!joined) {
+          this.leftSpace();
+        }
+      },
+      enteredSpace() {
+        self.isInSpace = true;
+      },
+      leftSpace() {
```
Daboo1998, line 99: maybe `onLeftSpace`?

Daboo1998 (changes requested):

PAKovalev (approved):

packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
```diff
@@ -30,9 +30,16 @@ const MeetingRoomPage: FC<PropsInterface> = ({onLeave}) => {
...
   useEffect(() => {
-    if (musicPlayerStore.musicPlayer.isPlaying) {
+    const wasPlaying: boolean = musicPlayerStore.musicPlayer.isPlaying;
+    if (wasPlaying && agoraStore.hasJoined) {
       musicPlayerStore.togglePlayback();
     }
+
+    return () => {
+      if (!musicPlayerStore.musicPlayer.isPlaying && wasPlaying) {
+        musicPlayerStore.togglePlayback();
```
Daboo1998, line 40: As my previous comment, I reccomment to change this, as its misleading. toggling means toggling between on and off, this seems to only switch on.

Daboo1998 (changes requested):

packages/app/src/scenes/meeting/pages/MeetingRoomPage/MeetingRoomPage.tsx
```diff
@@ -30,9 +30,16 @@ const MeetingRoomPage: FC<PropsInterface> = ({onLeave}) => {
   const location = useLocation();
 
   useEffect(() => {
-    if (musicPlayerStore.musicPlayer.isPlaying) {
+    const wasPlaying: boolean = musicPlayerStore.musicPlayer.isPlaying;
+    if (wasPlaying && agoraStore.hasJoined) {
```
dmitry-yudakov, line 34: maybe it's better to have it like `const wasPlayingBeforeJoining = player.isPlaying && agora.hasJoined` - because that's why we stop the player and we should consider agora snapshot state when trying to resume playing as well.

dmitry-yudakov (approved):
Generally approved with one small suggestion.

Daboo1998 (approved):

PR #346, opened 2022-10-17
feature(plugins): Optimised script loading to be loaded once, while plugins can be loaded multiple times

Daboo1998:
Scripts are now loaded seperately from the plugin components themselves.Multiple same plugins will reuse existing script


aminbksi (approved):

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -37,8 +37,10 @@ const CollaborationPluginPage: FC<PropsInterface> = ({pluginLoader}) => {
   }, []);
 
   useEffect(() => {
-    pluginLoader.init();
-  }, [pluginLoader]);
+    if (pluginsStore.canLoadPlugin(pluginLoader.name)) {
+      pluginLoader.loadPlugin();
+    }
+  }, [pluginLoader, pluginsStore]);
```
dmitry-yudakov, line 43: I don't really like adding an effect for this and the logic for loading. Maybe `pluginLoader.loadIfNeeded()` function that you simply call on every render would be enough.

Receiving `pluginLoader` as an argument here feels a bit unnatural, but not a huge problem. It could also be smt like:

```js
const {plugin, isLoading, isError} = pluginsStore.get(pluginId)
```
If it's not loaded it would do it and otherwise would return the cached plugin.

And looking at bigger picture - In future I think it should already be loaded at this stage - after rendering the tabs panel and so on so we'd just pass `plugin`.

packages/app/src/scenes/collaboration/Collaboration.tsx
```diff
@@ -69,8 +68,7 @@ const Collaboration: FC = () => {
   }, [agoraStore, history, rootStore, spaceId, t]);
 
   useEffect(() => {
-    pluginsStore.init();
-    console.info('fetched plugin list', pluginsStore.pluginLoaders.length);
+    pluginsStore.getSpacePlugins();
```
dmitry-yudakov, line 71: Reading this code I'd be wondering - what am I getting it for :) Maybe `startLoading()` or something would be clearer.

dmitry-yudakov (approved):
I think it's better that the original implementation but there are still few improvements I suggested.

packages/app/src/core/models/DynamicScriptLoader/DynamicScriptLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+      types.enumeration(Object.values(DynamicScriptLoaderStatusEnum)),
+      DynamicScriptLoaderStatusEnum.READY
+    ),
+    url: types.maybe(types.string),
+    name: types.maybe(types.string)
+  })
+  .volatile<{_scriptElement?: HTMLScriptElement}>((self) => ({
+    _scriptElement: undefined
+  }))
+  .actions((self) => ({
```
PAKovalev, line 17: why two actions sections?

packages/app/src/core/models/DynamicScriptLoader/DynamicScriptLoader.ts
```diff
@@ -0,0 +1,75 @@
...
+const DynamicScriptLoader = types
+  .model('DynamicScriptLoader', {
+    status: types.optional(
+      types.enumeration(Object.values(DynamicScriptLoaderStatusEnum)),
+      DynamicScriptLoaderStatusEnum.READY
+    ),
+    url: types.maybe(types.string),
+    name: types.maybe(types.string)
+  })
+  .volatile<{_scriptElement?: HTMLScriptElement}>((self) => ({
```
PAKovalev, line 14: You can use instead of volatile: 

.actions((self) => {
      _scriptElement: someType | undefined;

     return {
       actions: ....
    }
}

PAKovalev (approved):

PR #347, opened 2022-10-18
app: invite message content

aminbksi:
https://momentum.nifty.pm/l/eoeuIR8KQ3ur

grab a table toast compatible title/message


packages/app/src/scenes/unity/pages/UnityPage/UnityPage.tsx
```diff
@@ -88,10 +88,12 @@ const UnityPage: FC = () => {
     };
 
     toast.info(
-      <InviteToSpaceContent
+      <InvitationContent
         invitorName={invitorName}
         spaceName={spaceName}
-        joinToSpace={uiTypeId === appVariables.GAT_UI_TYPE_ID ? handleJoinTable : handleJoinSpace}
+        joinSpace={handleJoinSpace}
+        joinTable={handleJoinTable}
```
Daboo1998, line 95: Why split it? Maybe change the prop to one action called `join` or `accept`

packages/app/src/ui-kit/organisms/InvitationContent/InvitationContent.tsx
```diff
@@ -0,0 +1,46 @@
+import React, {FC} from 'react';
+import {useTranslation} from 'react-i18next';
+
+import {ToastContent} from 'ui-kit/molecules';
+
+interface PropsInterface {
+  spaceName?: string;
+  invitorName: string;
+  joinSpace?: () => void;
+  joinTable?: () => void;
```
Daboo1998, line 10: I would connect this as one action

Daboo1998 (changes requested):

packages/app/src/scenes/unity/pages/UnityPage/UnityPage.tsx
aminbksi: I thought it might be good since I am sending this prop `uiTypeId === appVariables.GAT_UI_TYPE_ID` as if it is a table invite or space invite

PAKovalev (approved):

packages/app/src/scenes/unity/pages/UnityPage/UnityPage.tsx
Daboo1998: hmm then what about `spaceName`? It does not match, cause you have `joinSpace` and `joinTable` but then you don't have `spaceName` **and** `tableName`

aminbksi: in table we don't have `tableName`, `spaceName` is only for invite to space, we filter it with `isTable` prop, the title and the onClick function, but it might be more reusable the way you suggested, I'll push it now

Daboo1998 (approved):

PR #351, opened 2022-10-19
fix(stream-chat): special characters +/- are not shown

dmitry-yudakov:
Turned out stream-chat treats messages as markdown by default and +/- are bullets symbols.

Their team suggested a way to overcome this and I made special support for single +/- characters. https://github.com/GetStream/stream-chat-react/issues/1816#issuecomment-1280380278

In other cases the standard behaviour is applied allowing to format text as bold and other markdown goodies

#DEV-100


jor-rit (commented):
hmm, lot of trouble for an obscure (and harmless) bug no normal user will even encounter...
But.... it is markdown, there are more special chars (*, #,) or strings ( ``` , ---)

It is more the 'problem' that we don't style it. If you input a `* foo`, it renders an <li>, but list item bullet does not show up.

For now I'd say it is more trouble then it is worth. Having markdown is kinda nice (as an easter egg I guess), so we don't want to disable it completely.  Fixing it with a ignorelist too much code.

Ohh, and you have to read this comment in 'raw' format, since github comments are also markdown, so have the same problem ;p

dmitry-yudakov:
@jor-rit Note, that I don't block markdown completely - just for single-character messages like "+" and "-" :) 
Something that should be done in markdown parser perhaps or stream chat IMO - I'm just a user trying to send "+" message to a friend :)

So it's a bit of a workaround plus the way to make more customisation potentially useful in future - finding the way to do what I did what a lot harder when it seems :))

jor-rit:
Yes, I know. But now you also have to check # and --- and ''' and ...
Whatever markdown renders and we show as empty bubble :/

dmitry-yudakov:
@jor-rit , @PAKovalev  ok, another approach :) instead of preventing markdown for some symbols like +/- I made the markdown bullets translated into <ul> visible in the chat messages.

PAKovalev (approved):

Daboo1998 (approved):

PR #354, opened 2022-10-20
Feature/cbr 41/user profile model

PAKovalev:
Small, but important improvements related to user models


packages/app/src/stores/SessionStore/SessionStore.ts
```diff
@@ -6,39 +6,34 @@ import {RequestModel, UserStatusEnum} from '@momentum-xyz/core';
...
 import {storage} from 'shared/services';
 import {api, FetchUserResponse} from 'api';
-import {UserProfileModel} from 'core/models';
+import {User} from 'core/models';
 import {LoginTypeEnum, StorageKeyEnum} from 'core/enums';
 import {guestProviderConfig, keycloakProviderConfig, web3ProviderConfig} from 'shared/auth';
 
 const SessionStore = types
   .model('SessionStore', {
+    userId: '',
```
Daboo1998, line 15: Do we need userId? Maybe lets make it a view instead that returns `user.id`? Or remove it

Daboo1998 (approved):

packages/app/src/stores/SessionStore/SessionStore.ts
PAKovalev: Let's leave userId, it is non-null value.

Daboo1998: But you asking response.id to it and also assign response to user, so why do we need this duplication?

PAKovalev: ahh, yeah, you are right.

packages/app/src/core/models/User/User.ts
```diff
@@ -2,57 +2,54 @@ import {types, Instance, flow} from 'mobx-state-tree';
...
       return self.inviteRequest.isDone;
     }),
+    // TODO: Move to another model
     setInvited(invited: boolean) {
       self.invited = invited;
     }
   }));
 
-export interface UserProfileModelInterface extends Instance<typeof UserProfileModel> {}
+export interface UserModelInterface extends Instance<typeof User> {}
```
aminbksi, line 53: do we need `Model` here? or `UserInterface` only?

PAKovalev: Model for all models so that to avoid misundestanding

aminbksi (approved):

dmitry-yudakov (approved):

PR #355, opened 2022-10-20
feature(SessionStore): actions were moved.

PAKovalev:


packages/app/src/core/models/User/User.ts
```diff
@@ -26,7 +26,7 @@ const User = types
   .views((self) => ({
     get avatarSrc(): string | undefined {
       return (
-        self.profile?.avatarHash &&
+        self.profile.avatarHash &&
```
aminbksi, line 29: I think you can drop it `self.profile.avatarHash &&`

aminbksi (approved):

packages/app/src/scenes/widgets/stores/ProfileMenuStore/ProfileMenuStore.ts
```diff
@@ -13,13 +15,17 @@ const ProfileMenuStore = types.compose(
       profileMenuDialog: types.optional(Dialog, {}),
       tokenRulesDialog: types.optional(Dialog, {}),
       tokenRuleReviewStore: types.optional(TokenRuleReviewStore, {isWorldList: true}),
-      tokenRulesStore: types.optional(TokenRulesStore, {})
+      tokenRulesStore: types.optional(TokenRulesStore, {}),
+      statusRequest: types.optional(RequestModel, {})
     })
     .actions((self) => ({
-      openProfileMenu() {
+      openProfileMenu(): void {
```
Daboo1998, line 22: Why? we never add void to actions and its not required.

Daboo1998 (approved):

dmitry-yudakov (approved):

packages/app/src/core/models/User/User.ts
PAKovalev: what? it is without `?` )

aminbksi: what? you're returning `string`, avatar hash, also I think you can drop `undefined`

PR #359, opened 2022-10-21
feature(fly with me): broadcasting

PAKovalev:
#FWM-50

Broadcasting buttons buttons must not trigger disengage  from flight.
[https://momentum.nifty.pm/l/RqKz9YT7TE](https://momentum.nifty.pm/l/RqKz9YT7TE)

### DIFF
--- a/packages/app/src/core/constants/routePaths.constants.ts
+++ b/packages/app/src/core/constants/routePaths.constants.ts
@@ -18,6 +18,7 @@ export const ROUTES = {
   },
   grabTable: '/:spaceId/table',
   collaboration: {
+    prefix: '/collaboration',
     base: '/collaboration/:spaceId',
     stageMode: `/collaboration/:spaceId/stage-mode`,
     stageModeControl: `/collaboration/:spaceId/stage-mode/control`,
--- a/packages/app/src/core/constants/routePaths.constants.ts
+++ b/packages/app/src/core/constants/routePaths.constants.ts
@@ -18,7 +18,7 @@ export const ROUTES = {
   },
   grabTable: '/:spaceId/table',
   collaboration: {
-    prefix: '/collaboration',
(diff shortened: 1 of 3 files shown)


packages/app/src/core/constants/routePaths.constants.ts
```diff
@@ -18,6 +18,7 @@ export const ROUTES = {
   },
   grabTable: '/:spaceId/table',
   collaboration: {
+    prefix: '/collaboration',
     base: '/collaboration/:spaceId',
```
Daboo1998, line 22: i would say base, and space instead. prefix sounds weird

Daboo1998 (changes requested):

dmitry-yudakov (approved):
It's ok this way, though I wonder if it shouldn't be tied to Don'tDisturb like the previous PR. Not sure if the passengers should go to DND mode though.

packages/app/src/scenes/widgets/pages/LiveStreamPIPWidget/LiveStreamPIPWidget.tsx
```diff
@@ -12,15 +12,17 @@ import {VideoPanel} from 'ui-kit';
...
+  const {mainStore, flightStore} = useStore();
   const {agoraStore, liveStreamStore} = mainStore;
+
   const history = useHistory();
   const {t} = useTranslation();
+  const location = useLocation();
+
+  const isFlightAround = useMemo(() => {
+    return !location.pathname.includes(ROUTES.collaboration.root);
+  }, [location.pathname]);
```
dmitry-yudakov, line 25: You know, I liked it more before - it was less spaghetti with this logic outside the component and given as the component property only. 
I'd only rename it to be more generic in terms of behaviour of this component.
Just an idea for the future.

PR #365, opened 2022-10-24
Feature/mir 17/implement use of generic attributes api in miro plugins

Daboo1998:


PAKovalev (approved):

aminbksi (approved):

packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
```diff
@@ -1,60 +1,72 @@
...
+      }
+    });
+  }, [setPluginSpaceStateSubField]);
 
-    return () => {
-      miroBoardStore.resetModel();
-    };
-  }, [miroBoardStore, request, spaceId]);
+  useEffect(() => {
+    init(['board']);
```
dmitry-yudakov, line 34: `init({ fields: ['board'] })` would be better here

packages/app/src/stores/MainStore/models/PluginsStore/PluginsStore.ts
```diff
@@ -1,54 +1,79 @@
...
+        api.pluginsRepository.getPluginsMetadata,
+        {plugin_uuids}
+      );
 
-      COLLABORATION_PLUGIN_LIST.forEach((plugin) => {
-        if (self.dynamicScriptsStore.containsLoaderWithName(plugin.name)) {
+      const pluginsOptions: GetPluginsOptionsResponse = yield self.pluginOptionsRequest.send(
+        api.pluginsRepository.getPluginsOptions,
+        {plugin_uuids}
+      );
```
dmitry-yudakov, line 46: Perhaps at least 

```js
const pluginData = yield Promise.all([
  self.pluginMetadataRequest.send(),
  self.pluginOptionsRequest.send()
])
```

packages/sdk/src/interfaces/pluginProps.interface.ts
```diff
@@ -1,19 +1,24 @@
...
 }
 
 export interface PluginTopBarActionInterface {
   main: () => JSX.Element | null;
 }
 
 export interface SpacePluginPropsInterface extends CorePluginPropsInterface {
+  init: (fields: string[]) => Promise<void>;
+  spacePluginState: PluginStateInterface;
+  setPluginSpaceStateSubField: <V>(
```
dmitry-yudakov, line 18: As a plugin developer I would be terrified seeing this `setPluginSpaceStateSubField` and would give up on the second camelCase step :)

In future I would go with firebase-styled `set/update` functions and hide all the subfields details in the implementation.

```
const updateSharedState = (path: string, data: any) => {
  if(typeof data === 'object') {
    // traverse data and make backend calls for setSubField(path, innerPath, value), perhaps with Promise.all
  } else {
    // call set subfield
  }
}
```

And further in future it would go to the backend

dmitry-yudakov (approved):
It's a good next step.

I suggested some small improvements and some bigger improvements. 

The bigger one is moving plugin state logic into separate plugin instance store - it would deal with loading/updating/monitoring the shared state thus making the plugin pluggable into more than just space.

packages/app/src/api/repositories/spaceRepository/spaceRepository.api.ts
```diff
@@ -0,0 +1,31 @@
...
+import {request} from 'api/request';
+
+import {spaceRepositoryEndpoints} from './spaceRepository.api.endpoints';
+import {
+  GetSpaceOptionsRequest,
+  GetSpaceOptionsResponse,
+  GetSpaceSubOptionRequest
+} from './spaceRepository.api.types';
+
+export const getSpaceOptions: RequestInterface<GetSpaceOptionsRequest, GetSpaceOptionsResponse> = (
```
PAKovalev, line 13: Move all of this to `spaceAttributeRepository` because `spaceAttributeRepository` is pure repo. It is used only by another repos. `spaceRepository` should have only own interfaces which will be transformed for `spaceAttributeRepository`.

PAKovalev (changes requested):

packages/sdk/src/interfaces/pluginProps.interface.ts
Daboo1998: Unfortunately Antons vision blocks us from doing it

packages/app/src/api/repositories/spaceRepository/spaceRepository.api.ts
Daboo1998: Well, options is not an attribute,  that's why its here

PAKovalev (approved):

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -38,6 +39,38 @@ const CollaborationPluginPage: FC<PropsInterface> = ({pluginLoader}) => {
     setActions(actions);
   }, []);
 
+  const setPluginState = async (field: string, value: unknown) => {
```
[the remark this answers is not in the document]
Daboo1998: it maps the setting state (synced with backend) to actuall api EP call
API calls should be done on the host (lines 47-55) and plugin would only call a simple setState function that does not know anything about API, just that you set some field to some value.

packages/app/src/api/repositories/spaceRepository/spaceRepository.api.ts
```diff
@@ -0,0 +1,31 @@
+import {RequestInterface} from '@momentum-xyz/core';
```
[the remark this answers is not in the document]
Daboo1998: Why? it contains both options and metadata, why would you call both option?

PAKovalev (changes requested):

packages/app/src/scenes/collaboration/Collaboration.tsx
```diff
@@ -173,13 +173,16 @@ const Collaboration: FC = () => {
       <Switch>
         {createRoutesByConfig(COLLABORATION_ROUTES)}
         {pluginsStore.spacePlugins.map((plugin) => {
+          const pluginState = pluginsStore.spacePluginStateList.find(
+            (pluginState) => pluginState.pluginId === plugin.id
+          );
```
dmitry-yudakov, line 178: perhaps `plugin` should contain `pluginState`? Or make spacePluginStateList an object, not array though I don't really see a point of hard separating the two of them?

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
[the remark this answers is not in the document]
dmitry-yudakov: It's not completely wrong, it's just a simplifying wrapper function.

I'm wondering though if `pluginState` cannot have simple `setState(field, val)` function and have the rest of the arguments stored in it? We might need to call it from other places in Space screen, not only CollaborationPluginPage - for `plugin_chat` for example.

It's not a huge problem to stay this way for now but good to add `useCallback` probably?

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -38,6 +39,38 @@ const CollaborationPluginPage: FC<PropsInterface> = ({pluginLoader}) => {
...
+    await pluginState?.init(
+      api.spaceAttributeRepository.getSpaceSubAttribute,
+      {
+        worldId: worldStore.worldId,
+        spaceId: space.id
+      },
+      pluginLoader.id,
+      options.fields
+    );
+  };
```
dmitry-yudakov, line 72: same here

packages/sdk/src/interfaces/pluginProps.interface.ts
```diff
@@ -1,11 +1,16 @@
...
 export interface CorePluginPropsInterface {
   theme: ThemeInterface;
   spaceId?: string;
   isSpaceAdmin: boolean;
-  request: AxiosInstance;
+
+  pluginState: PluginStateInterface;
+  init: (options: {fields: string[]}) => Promise<void>;
+  reload: () => Promise<void>;
+  setPluginState: <V>(key: string, value: V) => Promise<void>;
```
dmitry-yudakov, line 13: These could stay here for now but in the next version we'll probably have `sharedSpaceState` and `sharedWorldState` (or `sharedState`) - the space one would go to spacePluginProps interface then or now. 

I would also use the names that I mentioned, since they imply it's `shared`, because `plugin` state might be confusing.

dmitry-yudakov (changes requested):
I think it's good generally, made a few remarks as the one regarding the setPluginSpace should be addressed.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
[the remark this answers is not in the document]
Daboo1998: I cannot, cause otherwsie I cannot make it generic and type safe, at the same time

packages/app/src/scenes/collaboration/Collaboration.tsx
Daboo1998: Yes I agree, I moved it

packages/sdk/src/interfaces/pluginProps.interface.ts
Daboo1998: good point, will rename to sharedState

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
Daboo1998: would have to drop type safety for this

packages/app/src/api/repositories/spaceRepository/spaceRepository.api.ts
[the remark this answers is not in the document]
PAKovalev: Because spaceRepo will have own interfaces, it is FE data. 
spaceOptionRepository is common reusable stuff.

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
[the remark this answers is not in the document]
PAKovalev: Guys, `ApiPluginInterface `.

packages/app/src/api/repositories/spaceRepository/spaceRepository.api.ts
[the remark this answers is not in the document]
Daboo1998: What it is based on, I don't really know what you want spaceOptionRepository for? what is a space option? we don't have option but options and metadata

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
PAKovalev: Plugin should worry how to init ifself, not app.

Daboo1998: It does, we do not want to provide repositories though to the app...

PR #370, opened 2022-10-26
feature(Stage mode): Joining fixed

PAKovalev:
Joining space was fixed

### DIFF
--- a/packages/app/src/stores/MainStore/models/AgoraStore/AgoraStore.ts
+++ b/packages/app/src/stores/MainStore/models/AgoraStore/AgoraStore.ts
@@ -36,14 +36,10 @@ const AgoraStore = types
   // API Requests
   .actions((self) => ({
     getStageModeStatus: flow(function* (spaceId?: string) {
-      if (spaceId) {
-        return yield self.spaceIntegrationsRequest.send(
-          api.spaceIntegrationsRepository.fetchStageModeStatus,
-          {spaceId: spaceId ?? self.spaceId}
-        );
-      }
-
-      return undefined;
+      return yield self.spaceIntegrationsRequest.send(
+        api.spaceIntegrationsRepository.fetchStageModeStatus,
+        {spaceId: spaceId || self.spaceId || ''}
+      );
     })
   }))
   // Initializer


packages/app/src/stores/MainStore/models/AgoraStore/AgoraStore.ts
```diff
@@ -36,14 +36,10 @@ const AgoraStore = types
...
-          api.spaceIntegrationsRepository.fetchStageModeStatus,
-          {spaceId: spaceId ?? self.spaceId}
-        );
-      }
-
-      return undefined;
+      return yield self.spaceIntegrationsRequest.send(
+        api.spaceIntegrationsRepository.fetchStageModeStatus,
+        {spaceId: spaceId || self.spaceId || ''}
+      );
```
Daboo1998, line 42:
```suggestion

      if (!spaceId && !self.spaceId) {
        return undefined
      }
       
      return yield self.spaceIntegrationsRequest.send(
        api.spaceIntegrationsRepository.fetchStageModeStatus,
        {spaceId: spaceId || self.spaceId}
      );
```

Daboo1998 (changes requested):

packages/app/src/stores/MainStore/models/AgoraStore/AgoraStore.ts
PAKovalev: Why we have two spaceId at all?

PAKovalev: Eslint error will appear.

PAKovalev: We should refactor this. Not for now.

Daboo1998: I think for rejoining you have spaceId undefined and then you get spaceId that was defined before. If you join for first time, the spaceId gets defined

Daboo1998: But ok lets leave it, hope it won't produce bugs

Daboo1998 (approved):

aminbksi (approved):

PR #372, opened 2022-10-26
bugfix(app): space admin panel markups

aminbksi:
Scrollable panels in space admin


dmitry-yudakov (approved):
Seems safe, how much damage could 5 lines of css change make, right? :)

PAKovalev (approved):

PR #375, opened 2022-10-28
Feature/mir 17/fixes after pr close

Daboo1998:


packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -26,7 +26,12 @@ const PluginAttributesManager = types
 
       return value;
     }),
-    set: flow(function* <T>(worldId: string, spaceId: string, key: string, value: T) {
+    set: flow(function* <T>(
+      worldId: string,
+      spaceId: string,
+      key: string,
+      value: T extends undefined ? never : T
```
PAKovalev, line 33: Just `value: T | null`

dmitry-yudakov (approved):

packages/app/src/scenes/collaboration/pages/CollaborationPluginPage/CollaborationPluginPage.tsx
```diff
@@ -32,6 +32,10 @@ const CollaborationPluginPage: FC<PropsInterface> = ({pluginLoader}) => {
 
   pluginsStore.loadPluginIfNeeded(pluginLoader);
 
+  const api = useMemo(() => {
```
PAKovalev, line 35: What is it for? `attributesManager.getAPI(worldStore.worldId, space.id)` won't cause re-render.
If you do want to memoize this, just make view instead of action under store:

`.views((self) => ({
    getAPI(worldId: string, spaceId: string): APIInterface {
      return {
        get: (key) => self.get(worldId, spaceId, key),
        set: (key, value) => self.set(worldId, spaceId, key, value)
      };
    }`

packages/sdk/src/interfaces/api.interface.ts
```diff
@@ -1,4 +1,4 @@
 export interface APIInterface {
```
PAKovalev, line 1: `export interface APIInterface {
  get: <T>(key: string) => Promise<T | null>;
  set: <T>(key: string, value: T | null) => Promise<void>;
}`

it will be the same. Not?

PAKovalev (changes requested):

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
Daboo1998: but then, T can be undefined

packages/sdk/src/interfaces/api.interface.ts
Daboo1998: as `T` can be anything then it also can be null and undefined. `T | null` is pointless cause `T` without any specification what it is is the same as `T | null`

PAKovalev (approved):

PR #380, opened 2022-10-31
Plugin SDK - Host Emulator - phase 2

dmitry-yudakov:
Planned development:

- move MomentumRequired to HostEmulator and show automatically when on PROD
- separate world and space interfaces
- widgets panel
- use react-router

New command is added to `momentum-plugin` script - `start:plugin` - it runs the plugin with craco config and allows using it inside momentum. Currently it has problems with hot refresh and non-base roots after reload.

`momentum-plugin start` command (set as `start` script for plugin_miro) runs the dev server using react-scripts without custom craco staff - it works with hot refresh and reloads and generally is a more convenient options for plugin development in isolation.


packages/sdk/src/emulator/components/WorldEmulator/WorldEmulator.tsx
```diff
@@ -0,0 +1,38 @@
...
+  console.log('RENDER WorldEmulator', {plugin});
+
+  const history = useHistory();
+
+  return (
+    <styled.Container>
+      <Switch>
+        <Route exact path="/">
+          <div>
+            <button onClick={() => history.push(`/collaboration/${DUMMY_SPACE_ID}`)}>
```
PAKovalev, line 26: Lets create routes as const.

packages/sdk/src/emulator/components/SpaceEmulator/SpaceEmulator.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </styled.SpaceNav>
+      <styled.SpaceTabContainer>
+        <styled.SpaceTopBar>
+          <strong>Space / Plugin</strong>
+          {topBar}
+          <button onClick={() => onClose()}>X</button>
+        </styled.SpaceTopBar>
+        <styled.SpaceContent>
+          <Switch>
+            <Route path={`${path}/dashboard`}>
```
PAKovalev, line 43: Lets create routes as const.

PAKovalev (approved):

packages/sdk/bin/momentum-plugin.js
```diff
@@ -1,50 +1,72 @@
...
-        }
-
-        process.exit(child.status);
-        break;
+  case 'start:plugin':
+  case 'build':
+  case 'test': {
+    const nodeArgs = scriptIndex > 0 ? args.slice(0, scriptIndex) : [];
+    const configPath = require.resolve(`../craco.config.js`);
+    const command = script === 'start:plugin' ? 'start' : script;
```
Daboo1998, line 26: instead of `start:plugin` I would make it a parameter you pass to something like `momentum-plugin start --no-craco` or `momentum-plugin start --standalone` or something else like that (maybe something shorter)

packages/sdk/src/emulator/components/MomentumRequiredPage/MomentumRequiredPage.tsx
```diff
@@ -1,15 +1,16 @@
...
 import {PropsWithThemeInterface, Text} from '@momentum-xyz/ui-kit';
 
 import * as styled from './MomentumRequiredPage.styled';
 
 const MomentumRequiredPage: FC<PropsWithThemeInterface> = () => {
-  const {t} = useTranslation();
-
   return (
     <styled.Container>
-      <Text size="l" text={t('messages.momentumRequired')} weight="bold" />
```
Daboo1998, line 12: Why removed translation?

packages/sdk/src/emulator/components/SpaceEmulator/SpaceEmulator.tsx
```diff
@@ -0,0 +1,59 @@
...
+  plugin: PluginInterface;
+  onClose: () => void;
+}
+
+export const SpaceEmulator: FC<PropsInterface> = ({plugin, onClose}) => {
+  const {spaceId} = useParams<{spaceId: string}>();
+  console.log('RENDER SpaceEmulator', {plugin, spaceId});
+  const history = useHistory();
+
+  const [topBar = <span />, setTopBar] = useState<JSX.Element | null>(null);
```
Daboo1998, line 20: why `<span />` not as initial value in `useState` call?

Daboo1998 (changes requested):

packages/sdk/src/emulator/components/MomentumRequiredPage/MomentumRequiredPage.tsx
dmitry-yudakov: To spare the need of having the translation file for SDK and make it simpler.

packages/sdk/bin/momentum-plugin.js
dmitry-yudakov: It might mix with the underlying commands arguments - it's less ambiguous to use just just "command:specifier".
I also wanted the `start` command to work in standalone mode with hot refresh and so on - and `start:plugin` to be for the heavier and currently less functional version allowing to also plug it in local momentum.

Daboo1998: yes yes, I know I just don't like the `start:plugin` cause first of all we already have the word "plugin" in the script. I would suggest to look what arguments it can collide with and use a one that won't be colliding

Daboo1998 (approved):
I approve due to nature of the new setup

PR #384, opened 2022-11-01
Feature/mir 17/BE EPs fixes

Daboo1998:


packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -40,14 +46,24 @@ const PluginAttributesManager = types
...
+      } else {
+        yield self.setStateRequest.send(api.spaceAttributeRepository.setSpaceSubAttribute, {
+          worldId,
+          spaceId,
+          plugin_id: self.pluginId,
+          attribute_name: 'state',
+          sub_attribute_key: key,
+          value: value
+        });
+      }
```
dmitry-yudakov, line 66: Maybe reuse the data for both requests? Or directly `yield self.setStateRequest.send( value === null ? delEP : setEP, {...body}`.
But it's ok to stay like this too.

dmitry-yudakov (approved):

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -24,10 +24,16 @@ const PluginAttributesManager = types
         }
       );
 
-      if (value === undefined) {
+      if (!response) {
```
PAKovalev, line 27: one line:
`if (!response || !(key in response)) { 
   return null;
}`

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -40,14 +46,24 @@ const PluginAttributesManager = types
...
-        attribute_name: 'state',
-        sub_attribute_key: key,
-        value: value
-      });
+      if (value === null) {
+        yield self.setStateRequest.send(api.spaceAttributeRepository.deleteSpaceSubAttribute, {
+          worldId,
+          spaceId,
+          plugin_id: self.pluginId,
+          attribute_name: 'state',
```
PAKovalev, line 54: Move to enum value of 'attribute_name'. See `api/enums`

PAKovalev (changes requested):

PAKovalev (approved):

PR #385, opened 2022-11-02
feature(all): update axios to 0.22.0

Daboo1998:


PAKovalev (approved):

aminbksi (approved):

packages/core/src/models/Request/RequestModel.ts
```diff
@@ -110,6 +110,8 @@ const RequestModel = types
         if (error?.response?.status === UNAUTHORIZED_STATUS) {
           /* keycloak.login(); */
         }
+
+        return undefined;
```
dmitry-yudakov, line 114: Any reason to have this?

Daboo1998: yes, after updating axios suddenly I got that not all paths return a value. So I had to add that

PR #393, opened 2022-11-03
feature(screensharing): relay call fixed

PAKovalev:
Relay function was called all the time when user went back to ScreenShare tab.
It cases getting events from controller and then all users redirects back to ScreenShare tab.
Relay function is called once after starting screensharing now.

[https://momentum.nifty.pm/l/RSwkdjdWExy](https://momentum.nifty.pm/l/RSwkdjdWExy)

### DIFF
--- a/packages/app/src/scenes/collaboration/pages/ScreenSharePage/ScreenSharePage.tsx
+++ b/packages/app/src/scenes/collaboration/pages/ScreenSharePage/ScreenSharePage.tsx
@@ -36,8 +36,8 @@ const ScreenSharePage: FC = () => {
 
   const startScreenSharing = useCallback(async () => {
     const wasStarted: boolean = await agoraScreenShareStore.startScreenSharing(sessionStore.userId);
-    if (wasStarted) {
-      screenShareStore.relayScreenShare(space?.id ?? '');
+    if (wasStarted && space?.id) {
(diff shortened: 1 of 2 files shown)


Daboo1998 (approved):

packages/app/src/scenes/collaboration/pages/ScreenSharePage/ScreenSharePage.tsx
```diff
@@ -25,18 +25,21 @@ const ScreenSharePage: FC = () => {
...
     }
   }, [videoTrack, screenShareStore, sessionStore.userId]);
 
-  const startScreenSharing = useCallback(() => {
-    agoraScreenShareStore.startScreenSharing(sessionStore.userId);
-  }, [agoraScreenShareStore, sessionStore.userId]);
+  const startScreenSharing = useCallback(async () => {
+    const wasStarted: boolean = await agoraScreenShareStore.startScreenSharing(sessionStore.userId);
+    if (wasStarted) {
+      screenShareStore.relayScreenShare(space?.id ?? '');
```
dmitry-yudakov, line 40: it would be nicer to detect it by resolved Promise. This way works too, just less clear and more old-school :)

dmitry-yudakov (approved):
It's good. In future it would be better to have one screenShareStore instead of 2 :)

PR #395, opened 2022-11-03
Feature/mir 64/load config from be

Daboo1998:


packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -11,7 +11,8 @@ const PluginAttributesManager = types
     spaceId: types.string,
 
     getStateRequest: types.optional(RequestModel, {}),
-    setStateRequest: types.optional(RequestModel, {})
+    setStateRequest: types.optional(RequestModel, {}),
+    getConfigRequest: types.optional(RequestModel, {})
   })
   .actions((self) => ({
     get: flow(function* <T>(spaceId: string, key: string) {
```
PAKovalev, line 18: Btw: maybe `getState` will be better?

packages/plugin_miro/src/SpaceApp.tsx
```diff
@@ -12,20 +11,17 @@ import '@momentum-xyz/ui-kit/dist/themes/themes';
...
+const SpaceApp: FC<MiroPluginPropsInterface> = (props) => {
   const {theme} = props;
 
   useEffect(() => {
-    // @ts-ignore: window['env']
-    const env: AppConfigExtendedInterface | undefined = window['env'];
-
-    if (env) {
-      Object.entries(env).forEach((entry) => {
+    props.api.getConfig().then((config) => {
```
PAKovalev, line 18: It should be a part of store like:

`useEffect(() => {
 miroStore.init(api);
})`

PAKovalev (changes requested):

dmitry-yudakov (approved):

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
Daboo1998: I don't think so cause its not about getting state, but a particular value from it

packages/plugin_miro/src/stores/RootMiroStore.ts
```diff
@@ -0,0 +1,30 @@
...
+      miroBoardStore: types.optional(MiroBoardStore, {})
+    })
+  )
+  .actions((self) => ({
+    init: flow(function* () {
+      const config = yield self.api.getConfig();
+
+      Object.entries(config).forEach((entry) => {
+        const [key, value] = entry;
+        appVariables[key as keyof AppConfigInterface] = value as string;
```
dmitry-yudakov, line 23: It would be better to use the type of your actual config somehow without type assertion with `as`. You know what you expect, you use typescript generics in `api` but you don't really benefit from it.

It's not critical of course, just an observation.

packages/sdk/src/emulator/HostEmulator.tsx
```diff
@@ -1,11 +1,11 @@
 import {FC, useState} from 'react';
-import {PluginInterface} from 'interfaces';
+import {PluginConfigInterface, PluginInterface} from 'interfaces';
 
 import * as styled from './HostEmulator.styled';
 import {SpaceTabEmulator} from './components/';
 
-interface PropsInterface {
-  plugin: PluginInterface;
+interface PropsInterface<C extends PluginConfigInterface> {
```
dmitry-yudakov, line 7: Perhaps initializing type with default value like `<C extends PluginConfigInterface = PluginConfigInterface>` would work - to simplify using it later.

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
```diff
@@ -50,16 +52,33 @@ const PluginAttributesManager = types
           value
         }
       );
+    }),
+    getConfig: flow(function* <C extends GetSpaceAttributeResponse>() {
+      const response: GetSpaceAttributeResponse | undefined = yield self.getConfigRequest.send(
+        api.spaceAttributeRepository.getSpaceAttribute,
+        {
+          spaceId: appVariables.NODE_ID,
```
dmitry-yudakov, line 60: I'm not really sure why NODE_ID needs to be used for config, perhaps it's a temp thing if getting inherited space config is not available.

dmitry-yudakov (approved):

packages/plugin_miro/src/stores/RootMiroStore.ts
Daboo1998: ah yes forgot about that :)

packages/app/src/core/models/PluginAttributesManager/PluginAttributesManager.ts
Daboo1998: for now in plan is to use node config for plugins that is shared through the app. For example for Miro its MIRO_APP_ID and it should not differ. In future we can make two calls, one for node and one for config overrides for space, but that's maybe in the future

packages/sdk/src/emulator/HostEmulator.tsx
Daboo1998: Good idea

dmitry-yudakov (approved):

PR #398, opened 2022-11-04
Optimise git pre-commit hook by using eslint cache

dmitry-yudakov:
It should speed up the eslint pre-commit checks.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn run lint:errors
+yarn run lint:errors --cache
 yarn run prettier:fix
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn run lint:errors --cache
-yarn run prettier:fix
+yarn run prettier:fix --cache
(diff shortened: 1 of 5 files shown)


jor-rit (approved):

Daboo1998 (approved):
Nice

PAKovalev (approved):

.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn run lint:errors
-yarn run prettier:fix
+yarn run lint:errors --cache
```
PAKovalev, line 4: Just a comment. Does it work as expected ?

PAKovalev (approved):

.husky/pre-commit
dmitry-yudakov: It's complicated... :) Sometimes it's fast and other times still slow but a bit faster than before I'd say.

PAKovalev: Will husky prevent commit if there is eslint error ? )

dmitry-yudakov: Fortunately no because we wouldn't be able to commit any time soon :) 

But great question in fact - making running this linter completely useless :)

dmitry-yudakov: And running prettier:fix could perhaps be done before push interrupting it if there's any file not properly formatted - so we could correct it and commit and push again.

dmitry-yudakov: pre-commits is interrupted in case of eslint errors - I was wrong before

PR #400, opened 2022-11-04
Better handling of HTTP error 503

dmitry-yudakov:
Redirecting to maintenance page is pretty bad user experience - especially after momentum has been load and was working successfully. You cannot simply reload the page because of changed url and the existing state potentially working gets lost.

That's why I made a change - if the first request (`config/ui-client`) fails with 503 or something else, then we show Maintenance or SmtWentWrong error without changing url.

Further API requests with 503 are retried up to 3 times with incremental delay (1 sec, 2 sec, 4 sec) - configurable. If it still fails after retry, toast notification is shown.

It's implemented without using external libraries as they didn't play too well with our existing logic.


packages/app/src/scenes/App.tsx
```diff
@@ -34,10 +34,20 @@ const App: FC = () => {
   const {t} = useTranslation();
 
   useEffect(() => {
+    setAxiosRetry({
```
PAKovalev: Lets create a hook instead of useEffect

packages/app/src/api/request/request.ts
```diff
@@ -53,8 +53,48 @@ const getAccessToken = (): string => {
...
-export const setApiResponseHandlers = ({onResponse = responseHandler, onError = errorHandler}) => {
-  request.interceptors.response.use(onResponse, onError);
+type AxiosRequestConfigWithRetryType = AxiosRequestConfig & {momentumRetryCount?: number};
+
+export const setApiResponseHandlers = ({
+  onResponse = responseHandler,
+  onError = errorHandler,
+  maxRetries = 3,
+  retryDelayBase = 1000,
+  retryCodes = [503]
```
PAKovalev: Use existing constants.

packages/app/src/api/request/request.ts
```diff
@@ -53,8 +53,48 @@ const getAccessToken = (): string => {
...
 request.interceptors.request.use(requestHandler, errorHandler);
 
-export const setApiResponseHandlers = ({onResponse = responseHandler, onError = errorHandler}) => {
-  request.interceptors.response.use(onResponse, onError);
+type AxiosRequestConfigWithRetryType = AxiosRequestConfig & {momentumRetryCount?: number};
+
+export const setApiResponseHandlers = ({
+  onResponse = responseHandler,
+  onError = errorHandler,
+  maxRetries = 3,
```
PAKovalev: Let's move this to the env file.

packages/app/src/scenes/App.tsx
```diff
@@ -33,24 +33,39 @@ import 'react-toastify/dist/ReactToastify.css';
...
+    isConfigReady,
+    isError: isErrorLoadingConfig,
+    errorCode: configLoadingErrorCode
+  } = configStore;
 
   const {pathname} = useLocation<{pathname: string}>();
   const history = useHistory();
   const {t} = useTranslation();
 
   useEffect(() => {
```
PAKovalev, line 46: Lets create a hook under `shared/hooks`.

PAKovalev (approved):

packages/app/src/api/request/request.ts
```diff
@@ -34,7 +34,7 @@ const responseHandler = (response: AxiosResponse) => {
 };
 
 const errorHandler = (error: AxiosError) => {
-  throw error;
+  return Promise.reject(error) as unknown;
```
Daboo1998: why not change to async and then throw error?

packages/app/src/stores/ConfigStore/ConfigStore.ts
```diff
@@ -25,7 +25,7 @@ const ConfigStore = types
 
         Object.entries(appVariablesOverrides).forEach((entry) => {
           const [key, value] = entry;
-          appVariables[key as keyof AppConfigInterface] = value as any;
+          appVariables[key as keyof AppConfigInterface] = value as string;
```
Daboo1998, line 28: actually this is wrong. It should be 
```suggestion
          appVariables[key as keyof AppConfigExtendedInterface] = value as any;
```

as appVariables are of type `AppConfigExtendedInterface`

Daboo1998 (changes requested):

packages/app/src/stores/ConfigStore/ConfigStore.ts
dmitry-yudakov: It doesn't really matter - it's just casting of stuff and doesn't affect anything. And `any` will give us warning.

packages/app/src/scenes/App.tsx
dmitry-yudakov: I'll leave it as is for the moment. Lets discuss when you're back. I don't think it's good to hide all the logic of this effect in a hook, yet if it's just to spare useEffect+setApiResponseHandlers it's meaningless and only gives you pain of setting up the types and figuring out how to solve the issue of the inner deps list of the same useEffect.

packages/app/src/stores/ConfigStore/ConfigStore.ts
dmitry-yudakov: And it gives error `Type 'any' is not assignable to type 'never'.` if AppConfigExtendedInterface is used, whatever the hell it means. Lets just leave it as is ;)

packages/app/src/api/request/request.ts
dmitry-yudakov: It was needed for the first implementation, but now I just restored it to the original `throw error`.

packages/app/src/api/request/request.ts
dmitry-yudakov: Similar parameters like response timeout are in this file, so I put the new ones next to them.

Daboo1998 (approved):

PR #422, opened 2022-11-16
Feature/mir 27/change board

Daboo1998:
Implemented PosBus for plugins

### DIFF
--- a/packages/app/src/shared/services/unity/UnityService.tsx
+++ b/packages/app/src/shared/services/unity/UnityService.tsx
@@ -102,14 +102,14 @@ export class UnityService {
         msgJSON = '';
       }
       try {
-        PosBusService.handleRelayMessage(target, msgJSON);
+        PosBusService.main.handleRelayMessage(target, msgJSON);
       } catch (e) {
         console.error('Error in RelayMessage handler.', e);
       }
     });
 
     this.unityContext?.on('SimpleNotification', (kind: number, flag: number, message: string) => {
-      PosBusService.handleSimpleNotification(kind, flag, message);
+      PosBusService.main.handleSimpleNotification(kind, flag, message);
     });
   }
 
--- a/packages/app/src/shared/services/unity/UnityService.tsx
+++ b/packages/app/src/shared/services/unity/UnityService.tsx
@@ -102,14 +102,14 @@ export class UnityService {
         msgJSON = '';
       }
       try {
-        PosBusService.main.handleRelayMessage(target, msgJSON);
+        PosBusService.handleRelayMessage(target, msgJSON);
       } catch (e) {
         console.error('Error in RelayMessage handler.', e);
       }
     });
 
     this.unityContext?.on('SimpleNotification', (kind: number, flag: number, message: string) => {
-      PosBusService.main.handleSimpleNotification(kind, flag, message);
+      PosBusService.handleSimpleNotification(kind, flag, message);
     });
   }
 
--- a/packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
+++ b/packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
@@ -10,22 +10,12 @@ const MiroBoardPage: FC = () => {
   const {api, miroBoardStore} = useStore();
   const {board} = miroBoardStore;
   const {isAdmin, spaceId, renderTopBarActions, pluginApi} = useSpace();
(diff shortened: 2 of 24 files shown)


packages/app/src/shared/services/unity/UnityService.tsx
```diff
@@ -102,14 +102,14 @@ export class UnityService {
         msgJSON = '';
       }
       try {
-        PosBusService.handleRelayMessage(target, msgJSON);
+        PosBusService.main.handleRelayMessage(target, msgJSON);
```
dmitry-yudakov: The older version was simpler to use - why not hide this `main` inside the functions and leave them static?

packages/sdk/src/emulator/hooks/useAttributesEmulator.tsx
```diff
@@ -80,16 +84,24 @@ export const useAttributesEmulator = () => {
     return Promise.resolve();
   }, []);
```
aminbksi, line 85: what about dependencies here?

aminbksi (approved):

packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
```diff
@@ -9,12 +9,23 @@ import * as styled from './MiroBoardPage.styled';
...
 
   useEffect(() => {
     miroBoardStore.init(api);
     miroBoardStore.fetchBoard();
-  }, [api, miroBoardStore]);
+    subscribeToStateUsingTopic('miro-state');
+
+    return () => {
+      unsubscribeFromStateUsingTopic('miro-state');
+    };
```
dmitry-yudakov: I don't think you need separate subscribe/unsubscribe function - you can do it inside `useStateItemChange`. And the topic should probably be hidden and somehow be plugin id - not arbitrary string.

packages/sdk/src/interfaces/pluginApi.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+  deleteStateItem: (key: string) => Promise<null>;
+  getConfig: () => Promise<C>;
+  subscribeToStateUsingTopic: (topic: string) => void;
+  unsubscribeFromStateUsingTopic: (topic: string) => void;
+
+  useStateItemChange: <T = unknown>(
+    topic: string,
+    key: string,
+    callback: (value: T) => void
+  ) => void;
```
dmitry-yudakov: It'd be more convenient to have the hook return the value of type T and optionally provide a callback function. It could also read the initial value and return it - this way you would not have to keep the Miro board id in a store but instead just use the value that the hook returns to you.

It could be not `useStateItemChange` but `useStateItem` or `useSpaceState` (it's still not clear we're talking about current space/object, not whole world plugin state).

dmitry-yudakov (approved):
It can be merged this way, though I suggested some improvements.

packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
Daboo1998: Nope, as that's not how its implemented on BE, topics have to be specified

packages/sdk/src/emulator/hooks/useAttributesEmulator.tsx
Daboo1998: it will cause endless rerernders

packages/app/src/shared/services/unity/UnityService.tsx
Daboo1998: Good idea, let me see if it will work

packages/sdk/src/interfaces/pluginApi.interface.ts
Daboo1998: Lets also talk with Amin and Kovi (when he will be back) about it, cause I for now made it consistent with our usePosBusEvent() function

packages/plugin_miro/src/pages/MiroBoardPage/MiroBoardPage.tsx
Daboo1998: Also we don't want to listen to topics that are not supported, look at implementation on handleRelayMessage in PosBusService.ts

aminbksi (approved):

dmitry-yudakov (approved):

PR #424, opened 2022-11-17
Feature/google drive as plugin

Daboo1998:
Implemented google drive as plugin


packages/plugin_google_drive/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "plugin_google_drive",
+  "version": "0.1.0",
+  "private": true,
+  "port": 3002,
```
dmitry-yudakov, line 5: In future we better remove these ports from here and use `PORT=3002 yarn start`. Port 3001 should currently be passed in the plugin-momentum script.

packages/plugin_google_drive/public/index.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
```
dmitry-yudakov, line 1: We'll also need to see how to move these to the SDK.

dmitry-yudakov (approved):
I approve, It's ok considering the current architecture.

If we agree to have a hook like 

```js
const { value: doc, setValue } = useStateItem('doc');
```

it would probably spare us at least one store, or perhaps all of them.

packages/plugin_google_drive/package.json
Daboo1998: I think we should provide both ways, but rename port to devPort or something like that

aminbksi (approved):

packages/plugin_google_drive/public/index.html
Daboo1998: idk about that, cause the developer might want to change the index.html (craco or react-script, for example also do not hide it)

PR #435, opened 2022-11-24
Feature/load tiles

aminbksi:


packages/app/src/scenes/object/Object.tsx
```diff
@@ -1,37 +1,58 @@
...
       objectStore.resetModel();
     };
-  }, [rootStore, objectId, assetType, unityStore, objectStore]);
+  }, [history, objectId, objectStore, rootStore, unityStore]);
 
+  useEffect(() => {
+    if (assetType) {
+      history.push(generatePath(ROUTES.object.base, {objectId, assetType}));
+    }
+  }, [assetType, history, objectId]);
```
dmitry-yudakov: I wonder why do we need a route for this? Why not simply check the asset type and render the right page? It's not clear why they're "page" as well, but it seems we do use this term for lots of stuff :)

packages/app/src/scenes/object/pages/TextPage/TextPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+        </styled.HeaderElement>
+        <styled.HeaderElement className="right">
+          <styled.Button>
+            <SvgButton
+              iconName="close"
+              size="large"
+              isWhite
+              onClick={() => {
+                history.push(ROUTES.base);
+                console.info('????');
```
dmitry-yudakov: ????? :)

PAKovalev (approved):

dmitry-yudakov (approved):
Choosing the right renderer for the asset type can be simplified, apart from that it's ok.

dmitry-yudakov:
I think it's good now with the last change of using render function instead of routes

PR #449, opened 2022-11-29
Feature/screen share

aminbksi:


PAKovalev (approved):

packages/app/src/scenes/collaboration/pages/ScreenSharePage/ScreenSharePage.tsx
```diff
@@ -1,98 +1,99 @@
-import React, {FC, useCallback, useEffect} from 'react';
+import React, {FC} from 'react';
```
Daboo1998: I think instead of commenting the code out, lets leave the file as it was and just remove the link to it

packages/app/src/stores/MainStore/models/AgoraStore/AgoraScreenShareStore/AgoraScreenShareStore.ts
```diff
@@ -8,7 +8,7 @@ import AgoraRTC, {
 } from 'agora-rtc-sdk-ng';
 import {RequestModel, ResetModel} from '@momentum-xyz/core';
 
-import {api} from 'api';
+// import {api} from 'api';
 
 const TRACK_CONFIG: ScreenVideoTrackInitConfig = {
```
Daboo1998: I would not touch this file, just do not use the store

Daboo1998 (changes requested):

packages/app/src/stores/MainStore/models/AgoraStore/AgoraScreenShareStore/AgoraScreenShareStore.ts
aminbksi: how to resolve the api problem if we don't use the previous repository!

Daboo1998: You use the wrong file. You should change that in `app/src/stores/AgoraStore/AgoraScreenShareStore.ts` not in `app/src/stores/MainStore/models/AgoraStore/AgoraScreenShareStore/AgoraScreenShareStore.ts`

Daboo1998 (approved):

PR #537, opened 2022-12-12
Feature/3d connections

PAKovalev:


packages/app/src/stores/AuthStore/AuthStore.ts
```diff
@@ -60,6 +60,7 @@ const AuthStore = types.compose(
     .actions((self) => ({
       init(): void {
         self.token = getAccessToken();
+        self.wallet = getStoredWallet();
```
dmitry-yudakov, line 63: I think I'm checking it in couple of places to understand whether we're signed in but if token is invalid but wallet is stored, it might give some wrong indication. This change is a bit risky without taking into account the rest.

dmitry-yudakov (approved):

PR #550, opened 2022-12-19
Feature/std 62/3 d map

PAKovalev:


packages/map3d/src/components/Map3dCanvas/Map3dCanvas.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import {Map3dPropsInterface} from '../../interfaces';
+import {Map3dRender} from '../Map3dRender';
+
+import * as styled from './Map3dCanvas.styled';
+
+interface PropsInterface extends Map3dPropsInterface {}
+
+const Map3dCanvas: FC<PropsInterface> = (props) => {
+  const [isCanvasReady, setIsCanvasReady] = useState<boolean>(false);
```
Daboo1998: No need for genric if you set default value

packages/map3d/src/components/Map3dCanvas/Map3dCanvas.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const mapRef = useRef<HTMLCanvasElement | null>(null);
```
Daboo1998:
```suggestion
  const mapRef = useRef<HTMLCanvasElement>(null);
```
would also work

packages/map3d/src/interfaces/planet.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface PlanetInterface {
+  uuid: string;
+  owner: string;
+  name: string;
+  image?: string | undefined;
```
Daboo1998, line 5:
```suggestion
  image?: string;
```

why repeat yourself? :)

packages/map3d/src/interfaces/props.interface.ts
```diff
@@ -0,0 +1,9 @@
+import {PlanetInterface} from './planet.interface';
+
+export interface Map3dPropsInterface {
+  currentUserId: string | undefined;
```
Daboo1998: Are you sure not: 
```suggestion
  currentUserId?: string;
```
?

packages/map3d/src/interfaces/props.interface.ts
```diff
@@ -0,0 +1,9 @@
...
+  items: PlanetInterface[];
+  connections: Record<string, {id: string}[]>;
+  getImageUrl(imageUrlOrHash: string | undefined | null): string | null;
```
Daboo1998, line 7:
```suggestion
  getImageUrl(imageUrlOrHash?: string | null): string | null;
```
Not better?

Daboo1998 (approved):

aminbksi (approved):

packages/map3d/src/interfaces/props.interface.ts
PAKovalev: `imageUrlOrHash?` means that prop in not required.

packages/map3d/src/interfaces/planet.interface.ts
PAKovalev: It is two different cases. There is no field at all and prop is undefined.

Daboo1998: but it doesn't stop you from setting it as undefined explicitly 😊

```
image?: string;
```
is different from 
``` 
image: string | undefined
```
but not from 
```
image?: string | undefined;
```

packages/map3d/src/interfaces/props.interface.ts
Daboo1998: yes I know, That's why I am just asking if its not better 😉

unless you really need the user to explicitly say that its undefined (or the parameter passed)

PR #570, opened 2022-12-21
feat(app): search widget improvements

aminbksi:


packages/app/src/scenes/widgets/pages/SearchUsersWidget/components/organisms/OnlineUser/OnlineUser.tsx
```diff
@@ -0,0 +1,60 @@
...
+  currentUser: string;
+  onUserClick: (id: string) => void;
+  onTeleportUser: (id: string) => void;
+  onHighFiveUser: (id: string) => void;
+}
+
+const OnlineUser: FC<PropsInterface> = ({
+  user,
+  worldId,
+  currentUser,
```
dmitry-yudakov: it might be better to have `isCurrentUser`, `isAdmin`, `isCurrentWorld` flags passed from outside of this component; doesn't feel right to put the logic for figuring it out inside this component, some store is more suitable.

dmitry-yudakov (approved):

packages/app/src/scenes/widgets/pages/SearchUsersWidget/components/organisms/OnlineUser/OnlineUser.tsx
aminbksi: about `isAdmin`, I don't think it is a good phrase, because if they are not `guest`s it doesn't also mean they are admins. so I would keep it as it is.

PAKovalev (approved):

PR #571, opened 2022-12-21
3D Model preview

dmitry-yudakov:


packages/map3d/src/components/Model3dPreview/Model3dPreview.tsx
```diff
@@ -0,0 +1,119 @@
+import {useEffect, FC, useState, useRef} from 'react';
+import {Scene, DirectionalLight, AmbientLight, WebGLRenderer, PerspectiveCamera} from 'three';
+import {OrbitControls} from 'three/examples/jsm/controls/OrbitControls';
+import {GLTF, GLTFLoader} from 'three/examples/jsm/loaders/GLTFLoader.js';
+
+import * as styled from './Model3dPreview.styled';
+
+const createScene = (canvas: HTMLCanvasElement) => {
```
PAKovalev, line 22: It looks like a hook

dmitry-yudakov: It could be but it will all change when we're able to use react-three-fiber

FIZLIN (approved):
LGTM

PR #575, opened 2022-12-22
fix(auth): redirect to sign-in page if loading user profile fails

dmitry-yudakov:


PAKovalev (approved):

packages/app/src/scenes/AppAuth.tsx
```diff
@@ -9,7 +9,16 @@ const AppAuth: FC = ({children}) => {
 
   useEffect(() => {
     if (authStore.hasToken) {
-      sessionStore.loadUserProfile();
+      sessionStore
+        .loadUserProfile()
+        .catch((err) => {
```
PAKovalev, line 14: Not sure we need this change for now.

PAKovalev (changes requested):

PR #580, opened 2022-12-23
feat(get-tokens): Disable "Get tokens" while airdrop request is live

FIZLIN:


packages/app/src/scenes/widgets/pages/StakingWidget/components/templates/StakingDashboard/StakingDashboard.tsx
```diff
@@ -161,6 +162,7 @@ const StakingDashboard: FC = () => {
   };
   const onRequestAirdrop = () => {
     console.log('requestAirdrop', wallet);
+    setAirdropRequestLive(true); // TODO move this to actual request tracking?
```
dmitry-yudakov: Yep, we have it in `nftStore` - `requestingFundsStatus`. Search for usage example.

packages/app/src/scenes/widgets/pages/StakingWidget/components/templates/StakingDashboard/StakingDashboard.tsx
```diff
@@ -176,11 +178,13 @@ const StakingDashboard: FC = () => {
         );
         setCanRequestAirdrop(checkIfCanRequestAirdrop());
         setNextAvailableAirdropTime(getDateOfNextAllowedAirdrop());
+        setAirdropRequestLive(false);
       })
       .catch((err) => {
         console.log('requestAirdrop error', err);
         setCanRequestAirdrop(checkIfCanRequestAirdrop());
         setNextAvailableAirdropTime(getDateOfNextAllowedAirdrop());
```
dmitry-yudakov: it's a bit offtopic, but maybe we shouldn't set these in case of error?

dmitry-yudakov (changes requested):

packages/app/src/scenes/widgets/pages/StakingWidget/components/templates/StakingDashboard/StakingDashboard.tsx
FIZLIN: My reasoning was that a pletora of things can go wrong - the actual request might succeed and then something else might break in the flow. In that situation, if we don't do the extra check, the user will be able to click again and request another air drop. That's a _very_ edge case scenario, so I'm also fine with just removing it.

packages/app/src/scenes/widgets/pages/StakingWidget/components/templates/StakingDashboard/StakingDashboard.tsx
```diff
@@ -66,6 +67,7 @@ const StakingDashboard: FC = () => {
   const amountToken = amount / Math.pow(10, chainDecimals || 12);
 
   const [canRequestAirdrop, setCanRequestAirdrop] = useState<boolean>(checkIfCanRequestAirdrop());
+  const airdropRequestIsLive = requestingFundsStatus === 'pending';
```
dmitry-yudakov: Yep, the logic is correct now. 
This 'IsLive' is a bit confusing, I'd use 'InProgress' but it's not a big deal :)

packages/app/src/scenes/widgets/pages/StakingWidget/components/templates/StakingDashboard/StakingDashboard.tsx
dmitry-yudakov: But what if there's an actual failure - internet connection down for example - we won't do airdrop but won't allow the user to request it again :)

dmitry-yudakov (approved):
There few notes but in general it's ok to merge it.

PR #593, opened 2023-01-11
Feat/big number staking

FIZLIN:
Nifty task: https://momentum.nifty.pm/OUqUaZyl8i7Jl/task/ALP-57

### DIFF
--- a/packages/app/src/scenes/widgets/pages/ConnectWidget/components/templates/StakingForm/StakingForm.tsx
+++ b/packages/app/src/scenes/widgets/pages/ConnectWidget/components/templates/StakingForm/StakingForm.tsx
@@ -63,7 +63,6 @@ const StakingForm: FC<PropsInterface> = ({isGuest, nftItemId, onComplete}) => {
     addresses,
     accountOptions,
     nftItems,
-    chainDecimals,
     tokenSymbol
   } = nftStore;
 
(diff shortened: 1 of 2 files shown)


PAKovalev (approved):

packages/app/src/stores/NftStore/NftStore.ts
```diff
@@ -580,6 +588,9 @@ const NftStore = types
         stakedAtOthers: stakedAtOthersAddresses
       };
     }),
+    stakeBN: flow(function* (address: string, amount: BN, itemId: number) {
+      yield self.stake(address, amount.toNumber(), itemId);
+    }),
     stake: flow(function* (address: string, amount: number, itemId: number) {
```
dmitry-yudakov: I think you can change directly the type of `stake()` function and pass BN amount to `tx.stake.stake()`

packages/app/src/scenes/widgets/pages/ConnectWidget/components/templates/StakingForm/StakingForm.tsx
```diff
@@ -73,20 +74,25 @@ const StakingForm: FC<PropsInterface> = ({isGuest, nftItemId, onComplete}) => {
     : '';
 
   const [activeTab, setActiveTab] = useState<TabBarTabInterface>(tabBarTabs[0]);
-  const [amount, setAmount] = useState(DEFAULT_STAKING_AMOUNT);
-  const amountAtoms = amount * Math.pow(10, chainDecimals || 12);
+  const [amountString, setAmountString] = useState(DEFAULT_STAKING_AMOUNT.toString());
+  const [amountBN, setAmountBN] = useState(new BN(DEFAULT_STAKING_AMOUNT));
+  const [decimalAmount, setDecimalAmount] = useState(0);
```
dmitry-yudakov: I wonder if we can just have one string state and use BN for validation - I think they accept string constructor. We can keep state like `'1.25'` MOM and convert to MOM-cents just for validation and calling `stake()`

FIZLIN: The reason I decided to do it like this is because BN doesn't support decimal numbers.
I keep both the whole and decimal part so that I can convert it to BN and then validate.

packages/app/src/stores/NftStore/NftStore.ts
FIZLIN: ✅

packages/app/src/scenes/widgets/pages/ConnectWidget/components/templates/StakingForm/StakingForm.tsx
FIZLIN: I've made some changes to the logic as per our discussions, @dmitry-yudakov.

packages/app/src/stores/NftStore/NftStore.ts
```diff
@@ -207,6 +207,14 @@ const NftStore = types
...
       }
     },
+    canBeStakedBN(amount: BN): boolean {
+      try {
+        return self.balanceTransferrableBN.gte(amount);
+      } catch (err) {
+        console.error(err);
+        return false;
+      }
+    },
```
dmitry-yudakov: again, it's probably used in one place only and it's fine to change the original function :)

dmitry-yudakov (approved):
apart from one note it's fine

packages/app/src/stores/NftStore/NftStore.ts
FIZLIN: ✅

PR #600, opened 2023-01-13
Plugin SDK improvements and simplification, twitch/vimeo plugins

dmitry-yudakov:


packages/sdk/src/utils/UnityAutoPause.tsx
```diff
@@ -0,0 +1,15 @@
+import {FC, useEffect} from 'react';
+
+import {useUnityControl} from '../hooks';
+
+export const AutoPauseUnity: FC = () => {
```
PAKovalev, line 5: It is a hook. I should be named `useAutoPauseUnity`. Move to hooks also.

packages/sdk/src/utils/UnityAutoTakeKeyboardControl.tsx
```diff
@@ -0,0 +1,15 @@
+import {FC, useEffect} from 'react';
+
+import {useUnityControl} from '../hooks';
+
+export const UnityAutoTakeKeyboardControl: FC = () => {
```
PAKovalev, line 5: It is a hook. I should be named `useUnityAutoTakeKeyboardControl`. Move to hooks also.

packages/app/src/scenes/object/pages/ObjectPluginPage/ObjectPluginPage.tsx
```diff
@@ -56,25 +65,61 @@ const ObjectPluginPage: FC<PropsInterface> = ({plugin, pluginLoader, objectId})
...
+  return (
+    <ErrorBoundary errorMessage={t('errors.errorWhileLoadingPlugin')}>
+      <ObjectGlobalPropsContextProvider props={pluginProps}>
+        <PluginInnerWrapper pluginProps={pluginProps} plugin={plugin} pluginLoader={pluginLoader} />
+      </ObjectGlobalPropsContextProvider>
+    </ErrorBoundary>
+  );
+};
+
+const PluginInnerWrapper = ({
```
PAKovalev, line 92: Make a component under a `ObjectPluginPage`. One file === One component.

packages/plugin_vimeo/src/Plugin.tsx
```diff
@@ -0,0 +1,104 @@
...
+} from '@momentum-xyz/sdk';
+import {Button, Input, Text} from '@momentum-xyz/ui-kit';
+import '@momentum-xyz/ui-kit/dist/themes/themes';
+import {useState} from 'react';
+
+interface PluginStateInterface {
+  video?: string;
+}
+
+const usePlugin: UsePluginHookType = (props) => {
```
PAKovalev: Why do we have 2 plugins? One is for vimeo and second one for twitch?  These plugins are familiar. Secondly, we don't have plugin for youtube. My opinion, we should have only one plugin named `VideoViewer` (or something like this) which will be used for youtube, vimeo, twitch, etc. In order to support an additional video service - we just need to adapt existing plugin, we don't want to create and deploy a new one.

packages/plugin_vimeo/src/Plugin.tsx
```diff
@@ -0,0 +1,104 @@
...
+  ) : (
+    <Button onClick={() => setEditMode(true)} label="Change" />
+  );
+
+  const content = editMode ? (
+    <div style={{padding: '1em'}}>
+      <UnityAutoTakeKeyboardControl />
+      <Input
+        type="text"
+        label="Channel"
```
PAKovalev: i18n

packages/plugin_vimeo/src/Plugin.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <div
+      style={{
+        width: '100%',
+        height: '100%',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center'
+      }}
+    >
+      <Text text="Video not yet set" size="m" />
```
PAKovalev: i18n

packages/sdk/src/contexts/SpaceGlobalPropsContext.tsx
```diff
@@ -2,13 +2,11 @@ import React, {FC, useContext} from 'react';
...
 
-import {PluginConfigInterface, ObjectPluginPropsInterface} from '../interfaces';
+import {PluginPropsInterface} from '../interfaces';
 
 import {ThemeContextProvider} from './ThemeContext';
 
-export const ObjectGlobalPropsContext = createContext<
-  ObjectPluginPropsInterface<PluginConfigInterface>
->({
+export const ObjectGlobalPropsContext = createContext<PluginPropsInterface>({
```
PAKovalev, line 9: Take a look on file name. )

PAKovalev (changes requested):

packages/app/src/scenes/object/pages/ObjectPluginPage/ObjectPluginPage.tsx
dmitry-yudakov: It's more flexible this way, because you could put `< AutoPauseUnity/>` into conditionally rendered piece of code - not whole component. This way some video plugin may work while you're flying around in unity, but if you open edit form for changing video, it will take keyboard control.

packages/sdk/src/utils/UnityAutoTakeKeyboardControl.tsx
dmitry-yudakov: It's more flexible this way, because you could put `<UnityAutoTakeKeyboardControl/>` into conditionally rendered piece of code - not whole component. This way some video plugin may work while you're flying around in unity, but if you open edit form for changing video, it will take keyboard control.

packages/app/src/scenes/object/pages/ObjectPluginPage/ObjectPluginPage.tsx
dmitry-yudakov: Well, these aren't really 2 independent components - the second is more of an implementation detail required by the fact that I call a hook, consuming Context and I couldn't find a way to mix those in one component - and that's how we got 2 components but they're part of one logic and better stay together.

packages/plugin_vimeo/src/Plugin.tsx
dmitry-yudakov: yeah, it's a good idea

PR #630, opened 2023-02-03
fix(initial-airdrop): Made last airdrop based on curr user;

FIZLIN:


packages/app/src/stores/NftStore/NftStore.ts
```diff
@@ -671,15 +671,15 @@ const NftStore = types
         throw err;
       }
     }),
-    requestAirdrop: flow(function* (address: string) {
+    requestAirdrop: flow(function* (address: string, authWallet: string) {
```
dmitry-yudakov: `address` and `wallet` is the same in our case, so no need of another argument - just use `address`

dmitry-yudakov (approved):

PR #632, opened 2023-02-06
Skybox attributes

dmitry-yudakov:
https://momentum.nifty.pm/l/p4iEzUgzvrvW6_

Skybox list is read from node (not space) attribute right now - it allows to have one per node instead of one per space.

Current skybox is now kept in `active_skybox` attribute which is clearer than previous `skybox_custom`


FIZLIN (approved):
LGTM

PAKovalev (approved):

PR #647, opened 2023-02-13
feat(plugin-video): show invalid url, support additional params

dmitry-yudakov:
We should detect bad/unsupported url, show error input and not allow to save it. We should allow saving empty string though.

We should also pass Youtube timestamp into embed url, potentially similar attributes for other video types.

### DIFF
--- a/packages/plugin_video/src/Plugin.tsx
+++ b/packages/plugin_video/src/Plugin.tsx
@@ -22,28 +22,54 @@ interface PluginStateInterface extends Record<string, unknown> {
 const stateToQuery = (state: PluginStateInterface): string | null => {
   const {video_url, youtube_url, state: nestedState} = state;
   const url = video_url || youtube_url || nestedState?.video_url;
+
   // parse shared url and create embed url for youtube, twitch, vimeo
   if (url) {
-    if (url.includes('youtube.com')) {
-      const videoId = url.split('v=')[1].split('&')[0];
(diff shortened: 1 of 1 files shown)


PAKovalev (approved):

packages/plugin_video/src/Plugin.tsx
```diff
@@ -22,37 +22,64 @@ interface PluginStateInterface extends Record<string, unknown> {
...
-      const videoId = url.split('/')[3];
-      return `https://www.youtube.com/embed/${videoId}`;
-    } else if (url.includes('twitch.tv')) {
-      const videoId = url.split('/')[3];
-      if (/^\d+$/.test(videoId)) {
-        return `https://player.twitch.tv/?video=${videoId}&parent=${window.location.hostname}`;
-      } else {
-        return `https://player.twitch.tv/?channel=${videoId}&parent=${window.location.hostname}`;
+    try {
+      const [urlBase, paramsString] = url.split('?');
```
PAKovalev, line 29: Looks dirty. Maybe there is ready-to-use utils for working with this stuff.

FIZLIN (approved):

packages/plugin_video/src/Plugin.tsx
dmitry-yudakov: It could be better indeed, but it's a plugin and I'd prefer to keep it lean and simple enough.

PR #650, opened 2023-02-16
Feature/dev/storybook init

PAKovalev:


.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn run lint:errors --cache
-yarn run prettier:fix --cache
```
dmitry-yudakov, line 5: no cache? it speeds up things sometimes

packages/app/src/ui-kit/molecules/ToastMessage/ToastMessage.styled.ts
```diff
@@ -48,7 +48,7 @@ export const ToastContainerStyled = styled(ToastContainer)`
   }
 
   .Toastify__toast--info {
-    background: var(--new-blue-10);
+    background: #0517451a;
```
dmitry-yudakov, line 51: should we use them from the theme? or in the new ui-kit?

packages/app/package.json
```diff
@@ -17,6 +17,7 @@
     "@momentum-xyz/map3d": "^0.1.0",
     "@momentum-xyz/sdk": "^0.1.5",
     "@momentum-xyz/ui-kit": "^0.1.3",
+    "@momentum-xyz/ui-kit-storybook": "^0.1.0",
```
dmitry-yudakov, line 20: Not sure it's needed now. Unless we find a way to show storybook from some route in the app, like we do now.

packages/ui-kit-storybook/src/molecules/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const General = Template.bind({});
+General.args = {
+  size: 'large',
+  avatarSrc: 'https://picsum.photos/300'
+};
+
+export const IsOnline = Template.bind({});
+IsOnline.args = {
+  size: 'large',
+  avatarSrc: 'https://picsum.photos/300',
```
dmitry-yudakov, line 23: It seems avatarSrc could be reused here as part of template maybe? If I understand correctly the idea of templates.

Otherwise it seems it could be written with less boilerplate like in ts example here: https://storybook.js.org/docs/react/get-started/whats-a-story

Not a problem, just some ideas for optimisation.

dmitry-yudakov (approved):

FIZLIN (approved):

.husky/pre-commit
PAKovalev: Not really often. I don't see differences with cache and without )

packages/app/src/ui-kit/molecules/ToastMessage/ToastMessage.styled.ts
PAKovalev: Leave for later.

packages/app/package.json
PAKovalev: It is deployed as separate app.

packages/ui-kit-storybook/src/molecules/Avatar/Avatar.stories.tsx
PAKovalev: args === props of component

PR #675, opened 2023-03-13
Feature/dev 100/button

PAKovalev:


dmitry-yudakov (approved):
`round` and `ellipse` could be Button component's attributes, not separate components, but this way is ok too.

PR #680, opened 2023-03-15
Feat/babylon/loading tests

borislavstoev:
Some code snippets for Babylon, connected to the Emulator


packages/odyssey3d/src/babylon/ObjectHelper.ts
```diff
@@ -1,39 +1,146 @@
...
+
+    // Gizmo
+    this.gizmoManager = new GizmoManager(scene);
+    this.gizmoManager.clearGizmoOnEmptyPointerEvent = true;
+    //this.gizmoManager.positionGizmoEnabled = true;
+    //this.gizmoManager.rotationGizmoEnabled = true;
+    //this.gizmoManager.scaleGizmoEnabled = true;
+
+    // Testing
+    document.onkeydown = (e) => {
```
PAKovalev, line 104: It should be implemented as an useEffect. Skip this if it will be removed.

PAKovalev (approved):

packages/odyssey3d/src/babylon/ObjectHelper.ts
borislavstoev: Yes, I am going to use Babylon key detections instead and remove this later. There will be no document.onkeydown stuff:)

PR #684, opened 2023-03-15
Posbus + Babylon: Part 1

dmitry-yudakov:
Very early and ugly version


packages/app/src/scenes/App.tsx
```diff
@@ -64,6 +65,12 @@ const App: FC = () => {
     }
   }, [sessionStore, sessionStore.errorFetchingProfile]);
 
+  useEffect(() => {
+    if (sessionStore.token && sessionStore.user) {
+      PosBusService.init(sessionStore.token, sessionStore.user.id);
```
PAKovalev: `AppAuth` or `SessionStore` is better place for this stuff.

dmitry-yudakov: In `SessionStore` it's harder to determine the right moment the token and user are available somehow, but yeah, I can move it to `AppAuth`

PR #687, opened 2023-03-17
Feature/dev 100/voice chat

PAKovalev:


packages/ui-kit-storybook/src/static/styles/variables.css
```diff
@@ -2,6 +2,8 @@
   /** Primary colors **/
   --white: #ffffff;
   --black: #000000;
+  --brawn: #bb7851;
```
dmitry-yudakov: typo, should be brown

packages/ui-kit-storybook/src/molecules/VoiceUser/VoiceUser.styled.ts
```diff
@@ -0,0 +1,22 @@
+import styled from 'styled-components';
+import {rgba} from 'polished';
+
+const BROWN = '#bb7851';
+const YELLOW = '#eae71b';
```
dmitry-yudakov, line 5: it's set in 2 places, here and in css - maybe just here is good enough.

dmitry-yudakov (approved):

packages/ui-kit-storybook/src/static/styles/variables.css
PAKovalev: Yeah, fixed.

packages/ui-kit-storybook/src/molecules/VoiceUser/VoiceUser.styled.ts
PAKovalev: Let leave both places for now. Css-const will be used also, I think.

PR #696, opened 2023-03-23
Feat/babylon/loading tests

borislavstoev:


packages/odyssey3d/src/babylon/CameraHelper.ts
```diff
@@ -1,30 +1,55 @@
...
 import {ObjectHelper} from './ObjectHelper';
-//import * as BABYLON from '@babylonjs/core';
 
 export class CameraHelper {
-  //static camera: ArcRotateCamera | null = null;
-  static camera: FreeCamera | null = null;
+  static camera: UniversalCamera;
   static player: TransformNode;
+  static normalSpeed = 0.5;
+  static fastSpeed = 1;
```
dmitry-yudakov: These constants could be also defined outside the class - as `const` or `enum` with upper case letters perhaps - if their role is only giving a name to some constant value. But it's not critical and can stay this way.

dmitry-yudakov (approved):

packages/odyssey3d/src/babylon/CameraHelper.ts
borislavstoev: Thanks, addressed

PR #715, opened 2023-04-04
Feat/babylon/multiplayer

borislavstoev:


packages/odyssey3d/src/babylon/PlayerHelper.ts
```diff
@@ -9,18 +9,26 @@ import {
...
 const NORMAL_SPEED = 0.5;
 const FAST_SPEED = 1.5;
+const PLAYER_OFFSET = new Vector3(0, -0.5, 2);
 
-export class CameraHelper {
+export class PlayerHelper {
   static camera: UniversalCamera;
   static player: TransformNode;
+  static playerAssetId: string;
+  static playerMoveEvent: {unsubscribe: () => void} | undefined;
```
dmitry-yudakov: Do we need these actually? It seems they're not used.

packages/odyssey3d/src/babylon/ObjectHelper.ts
```diff
@@ -58,9 +59,13 @@ export class ObjectHelper {
...
         scene.pointerY,
         Matrix.Identity(),
-        CameraHelper.camera
+        PlayerHelper.camera
       );
 
+      ObjectHelper.lastClick = {
+        x: scene.pointerX,
+        y: scene.pointerY
+      };
```
dmitry-yudakov: Just `const lastClick = {...}`? Maybe no need of another static property?

dmitry-yudakov (approved):

packages/odyssey3d/src/babylon/PlayerHelper.ts
borislavstoev: Will need to unsubscribe from the playerMoveEvent when switching worlds eventually, so we will need that one in the recent future. 

The playerAssetId is something I started for a more distant future, where we allow players to customize the wisps in their world, but let's not get that far ahead for now

packages/odyssey3d/src/babylon/ObjectHelper.ts
borislavstoev: Yes, ty

PR #717, opened 2023-04-04
World3D: autohide gizmo, debounce UserMove message, send ObjectTransform to posbus

dmitry-yudakov:
Fix also potential issue with mutable callbacks passed to BabylonScene - they are stored in different helper classes but only the instance at the moment of onSceneReady calling. If the callback is changed afterwards, Babylon classes wouldn't know it.

To fix that I added `useMutableCallback` function storing the changing callback instance behind a constant instance - it can be used for storing it in external classes. It's also useful for the cases when we want to prevent some react components re-rendering due to potentially changeable arguments.


borislavstoev (commented):
lgtm

PAKovalev (approved):

PR #720, opened 2023-04-06
Universe3D: react-babylonjs Draft

dmitry-yudakov:


packages/odyssey3d/src/scenes/UniverseScene/UniverseScene.tsx
```diff
@@ -1,25 +1,150 @@
...
+      inst.actionManager.registerAction(
+        new ExecuteCodeAction(ActionManager.OnPickTrigger, onClick)
+      );
+    }
+  }, [inst, scene, onClick]);
+
+  return (
+    <mesh
+      name={`World ${info.id}`}
+      fromInstance={baseMesh}
```
jor-rit: hmm, baseMesh? Should this be the `inst` you created above?

packages/odyssey3d/src/scenes/UniverseScene/UniverseScene.tsx
```diff
@@ -1,25 +1,150 @@
...
+      position={position}
```
jor-rit, line 67: and position should then be set on the instance, otherwise the instancing doesn't have much use/effect.

jor-rit:
So the items.map to WorldOrb component is where I'm not sure what happens. Since my react knowledge is a bit out-of-date :)
And where i'm a bit skeptical about this approach and it reliance on the vdom.

But one or more items change, the react vdom figures out a diff... and recreates the changed components?
Fine it they are static, only created on initial mount.
The practical example here is: 5k orbs, but now 1 is added, the 'layout' engine recalculates positions for _all_ orbs.  So position would need to be an input property I guess (?). Meaning all WorldOrb components change (and start an animation to move towards the position) and react and Babylon start fighting for cpu/gpu time? :)

packages/odyssey3d/src/scenes/UniverseScene/UniverseScene.tsx
dmitry-yudakov: Yeah, it's something I don't yet understand and may need to dive into their code.
With `inst` in place of `baseMesh` it fails in runtime. 
And without `baseMesh.createInst()` earlier it will show only the last created sphere it seems.

dmitry-yudakov:
All good notes @jor-rit, I have the same questions :)

Here's my sandbox repo https://github.com/dmitry-yudakov/react_babylonjs_sandbox

I first tried with default approach of creating spheres with materials but it degrades pretty sharply from 10 to 100 to 500 to 5K. With `createInst` it's a lot better though it's unclear to me how this inst is used.

Last test generates 5K sphere instances and then adds 1 more every second.

It's not as bad as one would expect but it did lag a bit while flying around.

Adding `React.memo` to `WorldOrb` helped though sparing 5K+ re-renderings and it's pretty performant this way.

packages/odyssey3d/src/scenes/UniverseScene/UniverseScene.tsx
dmitry-yudakov: I think I fixed it and it makes more sense now :) See my last commit.
I'm using `instancedMesh` passing it `baseModel` as `source` attribute as it creates the instance itself.
https://brianzinn.github.io/react-babylonjs/examples/basic/instanced-mesh-hex

dmitry-yudakov:
@jor-rit My very last commit answers some of our questions :) 
Check also my sandbox repository - I'm testing with instantiating new orb every 100msec while dropping one of the old ones. Seemingly it doesn't leak and still provides a decent performance.

dmitry-yudakov:
I'll close this PR as it won't be merged soon and probably ever.

Maybe we could keep the branch though, just in case. If not, I have another sandbox project for the reference. 
@jor-rit - up to you.

jor-rit:
well, the PR is now forever in github :)
But yeah, I suggest you keep your own sandbox propject, since I myself, when the git branch list doesn't fit my old school 80x24 lines terminal, I start deleting 'stale' branches in github :stuck_out_tongue_winking_eye:

dmitry-yudakov:
I hope closed and unmerged PRs still go to some kind of heaven... RIP PR

PR #725, opened 2023-04-07
Feat/babylon/multiplayer

borislavstoev:


packages/app/src/shared/services/posBus/PosBusService.tsx
```diff
@@ -45,13 +45,13 @@ class PosBusService {
 
   private client: Client | null = null;
   private port: PosbusPort | null = null;
+  private static user: string;
```
dmitry-yudakov: `userId` would be a better name

packages/app/src/shared/services/posBus/PosBusService.tsx
```diff
@@ -141,9 +141,14 @@ class PosBusService {
 
       case MsgType.SET_WORLD: {
         console.log('Handle posbus set_world', data);
+        const {id, name, avatar, owner, avatar_3d_asset_id} = data;
+        console.log('emitting with id of: ' + PosBusService.user);
+        Event3dEmitter.emit(
+          'SetWorld',
+          {id, name, avatar, owner, avatar_3d_asset_id},
+          PosBusService.user
+        );
```
dmitry-yudakov: You can pass `data` directly, no need to recreate it.

```
Event3dEmitter.emit('SetWorld', data, PosBusService.userId)
```

packages/odyssey3d/src/babylon/PlayerHelper.ts
```diff
@@ -110,15 +140,121 @@ export class PlayerHelper {
...
+      const babylonUser = {
+        container: container,
+        userDefinition: user,
+        userInstance: instance
+      };
+      this.userMap.set(user.id, babylonUser);
+    }
+  }
+
+  static setUsersPosition(users: Odyssey3dUserTransformInterface[]) {
```
dmitry-yudakov: `userTransforms` would be a clearer name here

packages/odyssey3d/src/babylon/PlayerHelper.ts
```diff
@@ -110,15 +140,121 @@ export class PlayerHelper {
...
+    for (const user of users) {
+      if (user.id === this.playerId) {
+        return;
```
dmitry-yudakov: You probably want `continue` here, not `return` - it would return from function, breaking the loop and ignoring further transforms.

packages/odyssey3d/src/babylon/PlayerHelper.ts
```diff
@@ -110,15 +140,121 @@ export class PlayerHelper {
...
+      }
+      const userObj = this.userMap.get(user.id);
+
+      if (userObj) {
+        if (userObj.userInstance.rootNodes.length === 0) {
+          console.log('Cant set position, because the instance of user has no rootnode.');
+          return;
```
dmitry-yudakov: Also `continue` I'd say.

packages/odyssey3d/src/babylon/PlayerHelper.ts
```diff
@@ -110,15 +140,121 @@ export class PlayerHelper {
...
+        }
+
+        const transformNode = userObj.userInstance.rootNodes[0];
+
+        transformNode.position.x = user.transform.position.x;
+        transformNode.position.y = user.transform.position.y;
+        transformNode.position.z = user.transform.position.z;
```
dmitry-yudakov, line 230: Maybe it would be convenient to create a util function converting JS object `{x, y, z}` to Vector3 and use it like `transformNode.position = objToVec3(user.transform.position)`. Not sure performance wise how are the two approaches but I suspect they're pretty similar.

packages/odyssey3d/src/babylon/PlayerHelper.ts
borislavstoev: yes, I implemented it first for a single user. nice catch,ty

PR #733, opened 2023-04-11
Detect wallets installed

dmitry-yudakov:
We can now detect whether each wallet type is installed.

There's also small fix for the way of pre-setting Metamask when it's chosen - the old way forced showing the right extension but broke somehow the way it works and it wouldn't detect changing of account inside Metamask


PAKovalev (approved):

FIZLIN (approved):
LGTM

PR #736, opened 2023-04-13
feat: pass worlds/users arrays from universe store to universe emitter/babylon

dmitry-yudakov:


packages/core/src/services/event.emitter.ts
```diff
@@ -37,9 +37,9 @@ export const Event3dEmitter = new EventEmitter<Event3dType>();
 export type Event3dEmitterType = typeof Event3dEmitter;
 
 export type Universe3dType = {
-  WorldAdded: (world: WorldInfoInterface) => void;
+  WorldsAdded: (worlds: WorldInfoInterface[]) => void;
```
borislavstoev, line 40: eventually we are going to have both WorldsAdded and WorldAdded, right? Same for Users / User

dmitry-yudakov: I'm not sure yet

PR #752, opened 2023-04-21
Feat/babylon/universe generation

borislavstoev:


PAKovalev (approved):

packages/odyssey3d/src/babylon/UniverseBuilderHelper.ts
```diff
@@ -307,4 +314,225 @@ export class UniverseBuilderHelper {
...
+      id: 'Odyssey Particle',
+      capacity: 5,
+      disposeOnStop: false,
+      manualEmitCount: -1,
+      emitter: [-6, 12, 0],
+      particleEmitterType: {
+        type: 'PointParticleEmitter',
+        direction1: [0, 0, 0],
+        direction2: [0, 0, 0]
+      },
```
dmitry-yudakov: I usually avoid putting long sheets of stuff inside functions code as it's harder to read and understand. I'd put odyssey and spark objects outside the class or even separate file having `export const odysseyParticlesConf = {...` and import them from there.

dmitry-yudakov (approved):
It's ok with small remark that is not critical.

PR #785, opened 2023-04-28
High-five Bump effect

dmitry-yudakov:


packages/odyssey3d/src/scenes/BabylonScene/BabylonScene.tsx
```diff
@@ -109,6 +112,15 @@ const BabylonScene: FC<Odyssey3dPropsInterface> = ({events, ...callbacks}) => {
       events.on('ReceiveHighFive', (userId) => {
```
borislavstoev, line 112: Just making sure, 'SendHighFive' and 'ReceiveHighFive' will then become obsolete, right?

borislavstoev (approved):

packages/odyssey3d/src/scenes/BabylonScene/BabylonScene.tsx
dmitry-yudakov: No, they're still used for sending and receiving side but just for triggering the visual effect I guess.

borislavstoev: Ah okay, so just will not be used in Babylon logic then

PR #816, opened 2023-05-03
fix(babylon): rotation

dmitry-yudakov:
It seems the update function was using variables from the closure rather than taking fresh values. It's not clear though why it was like this for rotation only but worked for position and scale. To be investigated.


borislavstoev (approved):
weird it works like that, but I double tested it and this really seems to fix it

dmitry-yudakov:
Well, I can explain how it didn't work and how it works now (you can read about JS closure) - but I cannot explain how position and scale worked before :)

PR #846, opened 2023-05-05
Feature/cleaning up part1

PAKovalev:


packages/app/src/shared/hooks/useBlockchain.tsx
```diff
@@ -180,85 +179,3 @@ export const useBlockchain = ({requiredAccountAddress}: UseBlockchainPropsInterf
     getTokens
   };
 };
-
```
dmitry-yudakov, line 183: it's more convenient to keep them in one file as they're very tightly coupled... but ok :)

packages/ui-kit-storybook/src/atoms/WalletHash/WalletHash.tsx
```diff
@@ -1,19 +1,20 @@
...
+  icon?: IconNameType;
   hash: string;
 }
 
 const WalletHash: FC<WalletHashPropsInterface> = ({icon, hash}) => {
   return (
-    <styled.WalletContainer>
-      <IconSvg name="talisman" size="l" />
+    <styled.WalletContainer data-testid="WalletHash-test" className={cn(!icon && 'noIcon')}>
+      {icon && <IconSvg name="talisman" size="l" />}
```
dmitry-yudakov, line 17: `name={icon}` ?

dmitry-yudakov (approved):
Approved in general with 1 comment that should probably be addressed

packages/ui-kit-storybook/src/atoms/WalletHash/WalletHash.tsx
PAKovalev: There is a separate ticket for it )

PR #857, opened 2023-05-09
Refactor NftStore and remove several old widgets

dmitry-yudakov:
All obsolete stuff from NftStore are removed plus few other stores using it and few obsolete widgets using them.


FIZLIN (approved):
LGTM

PR #869, opened 2023-05-10
Creator: object deselect is explicit, prevent accidental selection switch

dmitry-yudakov:
The purpose is to avoid accidental selection of different objects while in creator mode - so if you want to change function of another object, you need first to deselect with X then select again with click

It's temp disabled for gizmo because it seems to have some internal logic making it switch the object on click.


packages/app/src/stores/UniverseStore/models/World3dStore/World3dStore.ts
```diff
@@ -57,6 +57,26 @@ const World3dStore = types
...
+  .actions((self) => ({
+    changeGizmoType(mode: GizmoTypeEnum) {
+      self.gizmoMode = cast(mode);
+      // TODO notify babylon
+      // UnityService.changeGizmoType(mode);
+    },
+    closeAndResetObjectMenu() {
+      console.log('closeAndResetObjectMenu', self.selectedObjectId);
+
+      const {creatorStore} = getRootStore(self);
```
PAKovalev, line 69: const {creatorStore, widgetManagerStore} = getRootStore(self);
I don't like mixing stores. I may be under component to have proper tests for stores in the future )

PAKovalev (approved):

packages/app/src/stores/UniverseStore/models/World3dStore/World3dStore.ts
dmitry-yudakov: I agree. I'm thinking about moving this stuff into creatorStore and not have to pass it through world3dStore at all. Not sure about widgetManager though.

borislavstoev (approved):

PR #874, opened 2023-05-12
Feature/object view

PAKovalev:


packages/app/src/scenes/App.routes.tsx
```diff
@@ -41,9 +40,5 @@ export const WORLD_ROUTES: RouteConfigInterface[] = [
     path: ROUTES.odyssey.base,
     main: () => <World2dPage />,
     exact: true
-  },
-  {
-    path: ROUTES.odyssey.object.root,
-    main: () => <Object />
```
dmitry-yudakov, line 47: Maybe we could keep this page and open the widget when it mounts? And change url to world base when it unmounts?

dmitry-yudakov (approved):
It would be useful to keep url to object. But it's not a blocker.

packages/app/src/scenes/App.routes.tsx
PAKovalev: It should be generic solution for all widgets include objects. Next step, I'd say.

PR #881, opened 2023-05-15
Migrate plugins to ui-kit-storybook

dmitry-yudakov:


packages/plugin_video/src/Plugin.tsx
```diff
@@ -120,31 +120,37 @@ const usePlugin: UsePluginHookType = (props) => {
...
-      />
-      <hr />
+      <div
+        style={{
+          textTransform: 'uppercase',
+          fontSize: 'var(--font-size-l)',
+          fontWeight: 600,
+          letterSpacing: '0.2em'
+        }}
+      >
```
dmitry-yudakov, line 144: It would be convenient to have this stuff as ui-kit component. @PAKovalev

packages/plugin_video/src/Plugin.tsx
```diff
@@ -120,31 +120,37 @@ const usePlugin: UsePluginHookType = (props) => {
...
+        {t('plugin_video.labels.embed')}
+      </div>
+      <p>
+        By embedding a video to this object; users will also be able to see this video played when
+        they select the object; regardless of its asset type.
+      </p>
+      <p>To embed a video; add the url to the video in the input field below.</p>
+      <div style={{backgroundColor: theme.accentText, height: 1}} />
```
dmitry-yudakov, line 152: and also this @PAKovalev

packages/plugin_video/src/Plugin.tsx
```diff
@@ -120,31 +120,37 @@ const usePlugin: UsePluginHookType = (props) => {
...
       <Input
-        type="text"
+        // type="text"
         // label={t('plugin_video.labels.videoUrl')}
         placeholder={t('plugin_video.messages.pasteUrl')}
-        autoFocus
+        // autoFocus
```
dmitry-yudakov, line 157: type, autoFocus, errorMessage are missing from Input but would be convenient to have @PAKovalev

PAKovalev (approved):

PR #902, opened 2023-05-16
fix: WIP General style fixes

FIZLIN:


packages/ui-kit/src/static/styles/variables.css
```diff
@@ -46,13 +46,18 @@
   --default-progress-bar: rgba(240, 145, 2, 1);
 
   /** Typography **/
-  --font-size-xxxl: 25px;
```
PAKovalev, line 49: it should be in the ui-kit-storybook.

PAKovalev (approved):

packages/ui-kit/src/static/styles/variables.css
FIZLIN: They are there, but now since we have both things, the storybook sizes were getting overwritten by the ui-kit ones.
Hence I updated the ui-kit ones to be like the storybook ones.

PR #907, opened 2023-05-18
Babylon: upgrade to v6

dmitry-yudakov:
AssetContainer instance rootNode is now Node type, not TransformNode, so it needs to be checked additionally.


borislavstoev (approved):
lgtm

PR #923, opened 2023-05-23
Feat/feature flag newsfeed

FIZLIN:


packages/app/src/api/constants/app.variables.ts
```diff
@@ -50,7 +50,10 @@ export const appVariables: AppConfigExtendedInterface = {
   NFT_ADMIN_ADDRESS: '',
   NFT_COLLECTION_ODYSSEY_ID: '',
   NODE_ID: '',
-  ODYSSEY_WORLD_ID: 'b8b104c4-b375-4933-8de8-29e0a1c67860'
+  ODYSSEY_WORLD_ID: 'b8b104c4-b375-4933-8de8-29e0a1c67860',
+  FEATURE_FLAGS: {
+    newsfeed: process.env.NODE_ENV === 'development'
+  }
```
dmitry-yudakov, line 56: I'm not sure we can handle such complex objects from the backend - it's better be string here and get parsed in `featureFlags.constants` if defined and not empty.

dmitry-yudakov: Turned out it can :)

dmitry-yudakov (approved):

PAKovalev (approved):

PR #972, opened 2023-07-03
Feature/dev 355/object sound

PAKovalev:


packages/core3d/src/babylon/ObjectHelper.ts
```diff
@@ -147,6 +169,50 @@ export class ObjectHelper {
...
+    const newTrack = newSoundInfo?.tracks.find((t: SoundItemInterface) => t.isActive);
+
+    // Only distance or volume was changed
+    if (newSoundInfo && prevObjectSound && prevObjectSound?.name === newTrack?.render_hash) {
+      prevObjectSound.updateOptions({
+        maxDistance: newSoundInfo.distance,
+        volume: newSoundInfo.volume / 100
+      });
+
+      this.appendObjectSound(obj.objectDefinition.id, prevObjectSound);
```
dmitry-yudakov, line 184: feels a bit overdesigned with these pop and append but no big deal :)

dmitry-yudakov (approved):

packages/core3d/src/babylon/ObjectHelper.ts
PAKovalev: Actually, it is because of waiting for spawning object.

PR #986, opened 2023-07-11
Hide faucet on prod

RamsesDijkstra:
I'm assuming this is enough? I was unable to test this, when i go to staking overview it complains that i have to switch to arbitrum network. Which the account i signed on with, is currently on.
So i've decided to just PR this, if it isn't sufficient then let me know

### DIFF
--- a/packages/app/src/scenes/widgets/pages/StakingViewWidget/components/MyWallet/MyWallet.tsx
+++ b/packages/app/src/scenes/widgets/pages/StakingViewWidget/components/MyWallet/MyWallet.tsx
@@ -8,6 +8,8 @@ import {WalletModelInterface} from 'core/models';
 import {useBlockchain} from 'shared/hooks';
 import {TokenSelector} from 'ui-kit';
 
+import {appVariables} from "../../../../../../api/constants";
+
 import * as styled from './MyWallet.styled';
 
 interface PropsInterface {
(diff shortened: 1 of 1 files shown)


packages/app/src/scenes/widgets/pages/StakingViewWidget/components/MyWallet/MyWallet.tsx
```diff
@@ -106,26 +108,31 @@ const MyWallet: FC<PropsInterface> = ({
           />
         </styled.RewardsContainer>
 
-        <styled.Title>Use Faucet</styled.Title>
-        <styled.AirdropContainer>
-          {canRequestAirdrop !== false ? (
+        {appVariables.CONTRACT_FAUCET_ADDRESS !== "" ?? (
```
dmitry-yudakov:
```suggestion
        {!!appVariables.CONTRACT_FAUCET_ADDRESS && (
```
This is a bit clearer to me :)
Using the double negation here means that if it's not `(empty string, null or undefined)`, we'll render the Faucet components.

packages/app/src/scenes/widgets/pages/StakingViewWidget/components/MyWallet/MyWallet.tsx
```diff
@@ -8,6 +8,8 @@ import {WalletModelInterface} from 'core/models';
 import {useBlockchain} from 'shared/hooks';
 import {TokenSelector} from 'ui-kit';
 
+import {appVariables} from "../../../../../../api/constants";
```
dmitry-yudakov: We use `baseUrl` in `tsconfig.json` so it should work as just `"api/constants"` without going to the X-th floor :)

dmitry-yudakov (changes requested):

jor-rit (approved):
yes, with @dmitry-yudakov comment, makes it more readable and 'safe' for whatever weird gets set (or not set) in the config

packages/app/src/scenes/widgets/pages/StakingViewWidget/components/MyWallet/MyWallet.tsx
RamsesDijkstra: seems like my ide was being naughty

dmitry-yudakov (approved):

PR #988, opened 2023-07-11
Claimable objects: 3D effect when unclaimed

dmitry-yudakov:
A special "transparent" effect can now be applied for Customisable objects that are not yet claimed - it's set after spawning such object and reset if unclaimed (custom content removed).

It's set to "none" when claimed by user - this removes the effect and restores the normal object view.

The "transparent" effect is similar to spawn preview - it's basically cloning the object and sets a special material all its meshes; the original object meshes are disabled. When clearing, it's the opposite - the clone is disposed and the original object is enabled.

The PR also contains the fix for deleting object attribute, which isn't really used here - if deleted, no posbus notification is received - so instead of deleting we're setting the attribute value to "none".

Attribute types from https://github.com/momentum-xyz/ubercontroller/pull/216 are needed here.

### DIFF
--- a/packages/app/src/api/repositories/spaceAttributeRepository/spaceAttributes.api.ts
+++ b/packages/app/src/api/repositories/spaceAttributeRepository/spaceAttributes.api.ts
@@ -61,12 +61,13 @@ export const deleteSpaceAttribute: RequestInterface<DeleteSpaceAttributeRequest,
 ) => {
   const {spaceId, plugin_id, attribute_name, ...restOptions} = options;
 
-  restOptions.data = {
-    plugin_id,
-    attribute_name
-  };
-
-  const url = generatePath(spaceAttributesRepositoryEndpoints().attributes, {spaceId});
+  const url =
+    generatePath(spaceAttributesRepositoryEndpoints().attributes, {spaceId}) +
+    '?' +
+    new URLSearchParams({
+      plugin_id,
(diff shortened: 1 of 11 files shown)


packages/app/src/api/repositories/spaceAttributeRepository/spaceAttributes.api.ts
```diff
@@ -61,12 +61,13 @@ export const deleteSpaceAttribute: RequestInterface<DeleteSpaceAttributeRequest,
...
   const {spaceId, plugin_id, attribute_name, ...restOptions} = options;
 
-  restOptions.data = {
-    plugin_id,
-    attribute_name
-  };
-
-  const url = generatePath(spaceAttributesRepositoryEndpoints().attributes, {spaceId});
+  const url =
+    generatePath(spaceAttributesRepositoryEndpoints().attributes, {spaceId}) +
```
PAKovalev: `const requestParams = {
    params: {plugin_id, ....},
    ...rest
  };`

PAKovalev (approved):

packages/app/src/api/repositories/spaceAttributeRepository/spaceAttributes.api.ts
dmitry-yudakov: Right, 10x :)

PR #1030, opened 2023-07-25
feat(blockchain): retrieve unstakes info from contract, claim unstaked

dmitry-yudakov:
We read the Staking contract `unstakes` mapping. 
Unfortunately it seems there's no convenient way of reading mapping address->UnstakeArray, that why I try to read them one by one starting from index 0 until I get an error.

The calculation for whether certain unstake is claimable is done by us using hardcoded 7 days period. Could probably be loaded from contract.

I sum total unstaked and total claimable (with unstake timestamp earlier than current timestamp minus 7 days) - in case claimable is greater than zero, "Claim unstaked" button is enabled.

Added also note about this 7 days unbonding period.

### DIFF
--- a/packages/app/src/shared/hooks/useBlockchain.tsx
+++ b/packages/app/src/shared/hooks/useBlockchain.tsx
@@ -2,6 +2,7 @@ import {useCallback, useMemo, useState} from 'react';
 import Web3 from 'web3';
 import BN from 'bn.js';
 import {TokenEnum} from '@momentum-xyz/core';
+import dayjs from 'dayjs';
 
 import {useStore} from 'shared/hooks';
 import {dummyWalletConf} from 'wallets';
@@ -20,6 +21,8 @@ import faucetABI from './contract_faucet.ABI.json';
 
 const DELAY_REFRESH_DATA_MS = 2000;
 
+export const UNBONDING_PERIOD_DAYS = 7;
+
 export interface UseBlockchainPropsInterface {
   requiredAccountAddress: string;
 }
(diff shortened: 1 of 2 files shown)


packages/app/src/shared/hooks/useBlockchain.tsx
```diff
@@ -136,6 +154,76 @@ export const useBlockchain = ({requiredAccountAddress}: UseBlockchainPropsInterf
...
   );
 
+  const getUnstakes = useCallback(
+    async (account: string, tokenKind: TokenEnum): Promise<UnbondingInfoInterface> => {
+      console.log('[UNSTAKES] Account ', account);
+      const latestUnblockedUnstakeTimestamp = dayjs().subtract(UNBONDING_PERIOD_DAYS, 'day').unix();
+      console.log('[UNSTAKES] latestUnblockedUnstakeTimestamp ', latestUnblockedUnstakeTimestamp);
+
+      const unstakes: ResolvedUnstakeInterface[] = [];
+      for (let i = 0; i < 1000; i++) {
```
jor-rit: I do get why... but 1000 magic number... :/
Currently unlikely 1000+ unstakes in process, but.... 
either add a comment documenting this limitation to fix in the future or use just a while loop ?

jor-rit (approved):

packages/app/src/shared/hooks/useBlockchain.tsx
dmitry-yudakov: Wondered about 0xFFFFFFFF or 0xdeadbeef but didn't want to stress potential readers :p anyway, removed it and added a comment

